PR #82, opened 2023-08-25
72 add recipient and list id notification config join

### LINKED ISSUE #72: Add recipient/list IDs to notification config

Currently, we are storing the recipient ids and recipient list ids for notification configs in the `configuration_data`. However, as every notification config will have these fields, it should be on the base NotificationConfig.

These should probably be two new join tables (notification_config_recipient & notification_config_recipient_list). They could then be added to the `NotificationConfigNode` via loaders in the graphql layer for the frontend.

Originally we thought to simply have `recipientIds` and `recipientListIds` as fields on the NotificationConfig, but this is a bit brittle (i.e. if a recipient is deleted, or what if we want to see that the notifications a recipient is subscribed to?) The join seems more robust/flexible.

lache-melvin:
Closes #72

# 👩🏻‍💻 What does this PR do?
 

Creates notification_config_recipient and notification_config_recipient_list entries when adding recipients/lists to a notification config. 

This makes things more flexible in terms of: handling if recipients/lists are deleted after added to a notification, if we ever were to display the notifications a recipient/list is currently set to receive.

As far as the frontend is concerned, recipientIds and recipientListIds are just fields on a NotificationConfig. They are attached to the GQL node via loaders. They are added/removed in the NotificationConfig create/update service layer methods. In the update, the current ids are diffed with the newly provided ones, so we only remove/add the necessary recipients/lists. Hopefully this keeps the logs cleaner, if/whenever we use them more. (Open to feedback on this implementation though!)

Unit tests added.

Create/update notification configs via the UI, adding/removing recipient/list members.

- I'm adding a new issue - this new join creates another foreign key constraint... we soft delete recipients, but not recipient lists or notification configs. We'll need to :)

## 📃 Documentation

_No user facing changes_


jmbrunskill (commented):
I think this is a useful reference, but don't think it adds a lot value at this stage in the project.
Let's shelve for now and come back to it, when we see it being useful...

PR #85, opened 2023-08-28
Respond to telegram bot messages

### LINKED ISSUE #32: Telegram bot should respond to direct messages with chatid

jmbrunskill:
Closes #32

# 👩🏻‍💻 What does this PR do?
 
The telegram bot now responds to commands such as /hello and /chatid

Also:

- [ ] Spawns telegram updates processing from a init method similar to Andrei's suggestion on previous PR
- [ ] Uses broadcast channel to allow multiple receivers for events

Chat with your bot asking `/help` or `/hello` or `/chatid`

I'm no longer aborting actix spawn threads. Seems to be working on on shutdown though. Is there some downside to this?

I'm not correctly able to tell in this method if the bot has just been added to a group to respond with a hello, so haven't implemented this yet.


backend/server/src/lib.rs
```diff
@@ -92,42 +92,33 @@ async fn run_server(
...
-        if let Some(telegram_token) = telegram_token {
-            if telegram_token.is_empty() {
-                log::error!("Telegram Client not configured");
-                return;
+    if let Some(telegram_token) = telegram_token {
+        let telegram_service = TelegramService::new(
+            TelegramClient::new(telegram_token),
+            config_settings.server.app_url.clone(),
+        );
+        let telegram_update_channel = telegram_service.init().await;
```
jmbrunskill, line 101: This creates a broadcast channel and spawns async threads to receive and process the events internally, while returning the channel for other subscribers to use...

backend/server/src/lib.rs
```diff
@@ -170,8 +161,8 @@ async fn run_server(
     };
 
     server_handle.stop(true).await;
-    telegram_update_handler.abort();
-    telegram_client_handle.abort();
+    // telegram_update_handler.abort();
```
jmbrunskill, line 164: Is this ok?
Because I'm, creating the spawns in another process and not passing it back here, it's not easy to abort.
Did a little bit of testing this seems to work fine? Maybe the oneshot restart thing doesn't work? Currently don't actually have that hooked up to an API...

backend/service/src/notification_config/tests/query.rs
```diff
@@ -168,7 +168,10 @@ mod notification_config_query_test {
                 &context,
                 None,
                 Some(NotificationConfigFilter::new().search("XXXX".to_string())),
-                None,
+                Some(Sort {
```
jmbrunskill, line 171: This is a fix for a flakey test

backend/telegram/src/service/mod.rs
```diff
@@ -1,690 +1,41 @@
...
-                "date": 1691536034,
-                "text": "This is a direct message message..."
-            }
+impl TelegramService {
+    pub fn new(client: TelegramClient, app_url: String) -> Self {
+        let (tx, _) = tokio::sync::broadcast::channel(10);
+        TelegramService {
+            client,
+            app_url,
+            updates_channel: tx, // we only need the transmission side because consumers just subscribe via this
```
andreievg, line 22: I am not sure why this is done, why not save rx vs subscribe ? (it seems straight forward to save rx vs subscribe, maybe it's just what I am used to)

backend/telegram/src/service/mod.rs
```diff
@@ -1,690 +1,41 @@
...
-        let last_update_id = handle_json_updates(updates, &tx).await;
-
-        // We quietly skip the bad update_id
-        assert_eq!(last_update_id, Some(794348052));
+    pub async fn init(self) -> Sender<crate::TelegramUpdate> {
+        // let app_url = self.app_url.clone();
+        let responder_channel = self.updates_channel.clone();
+        let responder_client = self.client.clone();
+        tokio::spawn(async move {
+            responder::handle_telegram_updates(&responder_client, &responder_channel).await;
```
andreievg, line 31: I think it's cleared if you move to async block, to also move it inside method (rather then accepting as reference)

backend/telegram/src/service/mod.rs
```diff
@@ -1,690 +1,41 @@
...
+        });
+
+        let sender_channel = self.updates_channel.clone();
+        tokio::spawn(async move {
+            poll_get_updates(&self.client, &sender_channel).await;
+        });
+        // Return the channel so consumers can subscribe
+        self.updates_channel.clone()
```
andreievg, line 39: don't think we need to clone, self is consumed by this method

backend/telegram/src/service/responder.rs
```diff
@@ -0,0 +1,58 @@
+// The responder should reply to the following...
+//
+// Direct messages to the bot
+// Messages to the bot with /hello or /chatid
+// Messages to the bot with /help
+
+use crate::{TelegramClient, TelegramUpdate};
+
+pub async fn handle_telegram_updates(
```
andreievg, line 9: I think one of the methods should be renamed.

andreievg (commented):
Had a quick look at things, I think there should be one pathway to handle telegram updates (one entry point to handle them). is there a reason to have two handle_telegram_updates methods ?

jmbrunskill:
> Had a quick look at things, I think there should be one pathway to handle telegram updates (one entry point to handle them). is there a reason to have two handle_telegram_updates methods ?

I guess there's a couple of reason, but perhaps not good ones...

1. Because the telegram stuff is in a separate crate, I didn't want to create a tight coupling between recipient service and telegram.
2. With channels this processing is happening concurrently, rather than one loop.

Maybe it's better to actually completely move the telegram crate into service?

I agree having one consumer for the messages is easier to understand. I quite like this approach, but maybe because it's cool, rather than easy to understand...

backend/telegram/src/service/mod.rs
jmbrunskill: The rx can only be used by 1 subscriber, so this api makes it clearer that new consumers need to subscribe, they can't just use the existing rx channel. I also have some issues with `SEND` or something I think, then realised this seems to be the pattern for broadcast...

backend/telegram/src/service/responder.rs
jmbrunskill: Done!

jmbrunskill:
> I agree having one consumer for the messages is easier to understand. I quite like this approach, but maybe because it's cool, rather than easy to understand...

I think in interest of time (both yours and mine) I'll merge this unless you feel strongly that the approach needs to change.
This is a nice to have feature not a mission critical one (at least for now)

PR #96, opened 2023-08-29
94 handle title template error

### LINKED ISSUE #94: Handle title template error
## Describe the bug

If the title template is missing, or has a render error blank title could be rendered

https://github.com/openmsupply/notify/blob/9238fc30a706aa219e3d1593c810ea980f7fbd95/backend/service/src/notification/enqueue.rs#L58

     `  log::error!("Failed to render notification template: {:?}", e);`

## To Reproduce

jmbrunskill:
Closes #94

# 👩🏻‍💻 What does this PR do?
 

Creates an errored notification if a notification title template can't be rendered

### DIFF
--- a/backend/service/src/notification/enqueue.rs
+++ b/backend/service/src/notification/enqueue.rs
@@ -53,21 +53,6 @@ pub fn create_notification_events(
         let mut template_context: serde_json::Value = notification.template_data.clone();
         template_context["recipient"] = serde_json::to_value(recipient.clone()).unwrap_or_default();
 
-        let title = match notification.title_template_name.clone() {
-            Some(title_template_name) => Some(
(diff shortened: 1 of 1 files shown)


backend/service/src/notification/enqueue.rs
```diff
@@ -78,23 +63,57 @@ pub fn create_notification_events(
...
+                            error_message: Some(format!("{:?}", e)),
+                            ..base_row
+                        }
+                    }
+                }
+            }
+            None => base_row,
+        };
+
+        let notification_queue_row = match base_row_with_title.status {
```
jmbrunskill, line 95: I don't like how nested this ends up being. Maybe multiple `repo.insert_one` records is better than this base_row approach.

PR #97, opened 2023-08-29
Code to run sql query from graphql request (Here be Dragons!)

### LINKED ISSUE #98: Ability to query recipients from Datasource
As a configurator, I want to be able to get a list of recipients via an SQL query.

This feature should be implemented as a Proof of concept, and something we can improve on an extend in future.

jmbrunskill:
Closes #98

# 👩🏻‍💻 What does this PR do?
 

Kapture 2023-08-30 at 16 26 03

Would be good to test some more complex queries...

**Focus for review should be on things like approach t connectivity to the database using a connection pool.**

There's a number of things we need to extend here.
For example...

1. We plan to use a recipient specific graphql request and parse the sql query result in the backend (needed to create notifications)
2. We plan to parameterise these queries somehow in future (possibly having a template driven approach)

Happy to have input on those but might be a separate PR, unless we think this is just too dangerous to merge (which it technically probably is if we had bad actors interacting with the database.

### DIFF
--- a/backend/datasource/src/connection_pool.rs
+++ b/backend/datasource/src/connection_pool.rs
@@ -4,12 +4,18 @@ use diesel::r2d2::Pool;
 
 use crate::PostgresSettings;
 
-pub fn get_datasource_pool(settings: &PostgresSettings) -> Pool<ConnectionManager<PgConnection>> {
+pub struct DatasourcePool {
+    pub pool: Pool<ConnectionManager<PgConnection>>,
+}
+
+pub fn get_datasource_pool(settings: &PostgresSettings) -> DatasourcePool {
     let url = settings.connection_string();
     let manager = ConnectionManager::<PgConnection>::new(url);
 
-    Pool::builder()
+    let pool = Pool::builder()
         .min_idle(Some(1))
         .build(manager)
-        .expect("Could not create datasource connection pool")
+        .expect("Could not create datasource connection pool");
+
+    DatasourcePool { pool }
 }
--- a/backend/service/src/service_provider.rs
+++ b/backend/service/src/service_provider.rs
@@ -5,6 +5,7 @@ use telegram::TelegramClient;
 
 use crate::{
     auth::{AuthService, AuthServiceTrait},
+    datasource::{DatasourceService, DatasourceServiceTrait},
     email::{EmailService, EmailServiceTrait},
     notification_config::{NotificationConfigService, NotificationConfigServiceTrait},
     recipient::{RecipientService, RecipientServiceTrait},
@@ -15,9 +16,9 @@ use crate::{
 
 pub struct ServiceProvider {
     pub connection_manager: StorageConnectionManager,
+    pub datasource_service: Box<dyn DatasourceServiceTrait>,
     pub email_service: Box<dyn EmailServiceTrait>,
     pub validation_service: Box<dyn AuthServiceTrait>,
-    pub general_service: Box<dyn GeneralServiceTrait>,
     pub user_account_service: Box<dyn UserAccountServiceTrait>,
     pub notification_config_service: Box<dyn NotificationConfigServiceTrait>,
     pub recipient_service: Box<dyn RecipientServiceTrait>,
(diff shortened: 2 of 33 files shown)


backend/service/src/service_provider.rs
```diff
@@ -73,11 +74,18 @@ impl ServiceProvider {
             None => None,
         };
 
+        //TODO Rearrange this so we're not using port number to determine if we want to mock...
```
jmbrunskill, line 77: Keen for ideas on how to do this, or avoid this problem...

frontend/packages/system/src/Recipients/RecipientLists/DetailView.tsx
```diff
@@ -139,13 +171,36 @@ export const DetailView = () => {
...
-                </LoadingButton>
+                <>
+                  {list?.sqlQuery && (
+                    <LoadingButton
+                      variant="outlined"
+                      isLoading={sqlIsLoading}
+                      startIcon={<ZapIcon />}
+                      onClick={() => {
+                        runSqlQuery(
+                          list?.sqlQuery ?? "SELECT 'SQL QUERY MISSING' as name"
```
jmbrunskill, line 182: This case shouldn't ever be hit, but TS doesn't seem to know that

frontend/packages/system/src/Recipients/api/hooks/useSQLRecipients.ts
```diff
@@ -0,0 +1,38 @@
...
+import { getSdk } from '../operations.generated';
+
+export interface BasicRecipientRow {
+  id: string;
+  name: string;
+  notificationType: string;
+  toAddress: string;
+}
+
+interface LcRecipientRow {
```
jmbrunskill, line 11: For some reason capitalisation seems to be lost in processing to json or something?
This will be refactored to use a specific graphql request for recipientlists..

frontend/packages/system/src/Recipients/api/operations.graphql
```diff
@@ -1,3 +1,10 @@
+fragment BasicRecipientRow on RecipientNode {
```
jmbrunskill, line 1: I think we'll need this for the new sqlendpoint, but isn't actually used right now...

frontend/packages/system/src/Recipients/RecipientLists/DetailView.tsx
clemens-msupply: The initial `list?.sqlQuery` check is not in the current arrow function. At the time onClick is called `list?.sqlQuery` could have been set to `undefined` in the meantime...

You could create a local `const sqlQuery: string` variable if you want to avoid the check...

backend/datasource/src/connection_pool.rs
```diff
@@ -4,12 +4,18 @@ use diesel::r2d2::Pool;
 
 use crate::PostgresSettings;
 
-pub fn get_datasource_pool(settings: &PostgresSettings) -> Pool<ConnectionManager<PgConnection>> {
+pub struct DatasourcePool {
+    pub pool: Pool<ConnectionManager<PgConnection>>,
+}
+
+pub fn get_datasource_pool(settings: &PostgresSettings) -> DatasourcePool {
```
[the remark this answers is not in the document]
jmbrunskill: There's an SQL Database which stores the configuration, recipients, users, etc. this is the `repository` and then a `datasource` which is the connection to a separate postgres database (Could be conforma, msupply, tumanu, what ever)

At the moment I'm assuming we have just 1 datasource, but I can see a potential to have more than one, but wanted to keep it simple for now.

Having it as a separate crate provides are separation of concerns.
Happy to have other ideas proposed though.

Datasource vs repostiory maybe isn't the best naming but I haven't come up with anything better yet.

frontend/packages/system/src/Recipients/RecipientLists/DetailView.tsx
jmbrunskill: Ah thanks, that's a good idea to add a local check.

frontend/packages/system/src/Recipients/RecipientLists/DetailView.tsx
```diff
@@ -62,6 +74,19 @@ export const DetailView = () => {
     }
   }, [suffix, list]);
 
+  useEffect(() => {
+    if (list?.sqlQuery) {
+      runSqlQuery(list?.sqlQuery)
```
clemens-msupply, line 79: Wonder if the server should run the query? i.e. you just call getRecipients and the server figures out where to get them from... That would reduce the attack surface, i.e. only server can see and execute sqlQueries and less likely FE injects some bad sql...

clemens-msupply (approved):
Only a quick review, but looks good!

Regarding DataSources, without too much thought on it (can chat tomorrow), I wonder if you can have multiple DataSources and types. For example, you can configure a `RecipientExtractor::new(source: PostgresDataSource)`, which can be configured with a custom sql query, i.e. store the query with the data source entry...

frontend/packages/system/src/Recipients/RecipientLists/DetailView.tsx
jmbrunskill: Yes I think so, here's a ticket I create for this... https://github.com/openmsupply/notify/issues/99

Issue #92, opened 2023-08-29
backoff for notification retries


jmbrunskill:
## Is your feature request related to a problem? Please describe
If a notification can't be sent with a temporary error, we should retry sending, but should wait longer between each attempt.

## Describe the solution you'd like

Suggest something like

1. Try to send immediately
2. Wait 1 Minute try again
3. Wait 15 Minutes try again
4. Wait 1 Hour try again

## Describe alternatives you've considered

We could have more attempts and/or change the logic abit, I'm not too precious about it.
I think anything that can't be sent say with in 12 hours definitely shouldn't be sent, but maybe we should try again within 2-4 hours?

PR #117, opened 2023-09-08
Email Sending

### LINKED ISSUE #91: Implement email notification sending

jmbrunskill:
Closes #91

# 👩🏻‍💻 What does this PR do?
 

- Refactors email queue (Used for sending password reset etc)
- Enabled email sending for notifications
- Test case in coldchain which can be run with 
`cargo test --features email-tests,telegram-tests  --package coldchain --lib -- tests::test_send_high_temperature_alert --exact --nocapture`

You'll need to run `mailhog` and setup your telegram env 

I plan to do exponential back off at some stage.
https://github.com/openmsupply/notify/issues/92

### DIFF
--- a/backend/coldchain/src/lib.rs
+++ b/backend/coldchain/src/lib.rs
@@ -3,7 +3,7 @@ use serde::Serialize;
 use service::{
     notification::{
         self,
-        enqueue::{create_notification_events, NotificationContext, NotificationRecipient},
+        enqueue::{create_notification_events, NotificationContext, NotificationTarget},
     },
     service_provider::ServiceContext,
 };
@@ -41,7 +41,7 @@ pub struct TemperatureAlert {
 pub async fn send_high_temperature_alert_telegram(
     ctx: &ServiceContext,
     alert: TemperatureAlert,
-    recipients: Vec<NotificationRecipient>,
+    recipients: Vec<NotificationTarget>,
 ) -> Result<(), notification::NotificationServiceError> {
     let notification = NotificationContext {
         title_template_name: Some("coldchain/telegram/temperature_title.html".to_string()),
@@ -72,7 +72,7 @@ mod tests {
     use service::test_utils::telegram_test::get_default_telegram_chat_id;
     use service::test_utils::telegram_test::send_test_notifications;
 
-    use crate::notification::enqueue::NotificationRecipient;
+    use crate::notification::enqueue::NotificationTarget;
     use std::str::FromStr;
 
     use service::service_provider::ServiceContext;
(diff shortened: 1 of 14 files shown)


backend/coldchain/src/lib.rs
```diff
@@ -102,32 +103,46 @@ mod tests {
             temperature: 10.12345,
         };
 
-        let recipient = NotificationRecipient {
+        let recipient1 = NotificationRecipient {
```
clemens-msupply, line 106: Nit: is there a better name for `NotificationRecpient`? sounds like an entry from the recipients list but I believe its more like `NotificationTarget` / `NotificationReceiver`?

backend/service/src/email/send.rs
```diff
@@ -0,0 +1,57 @@
+/*
+    This file defines the functionality needed to send an email to a to address (provided as a string)
+    with a subject (provided as a string) and a body (provided as a string).
+    It should return an error format with either a permanent error (which should be logged and not retried)
+    or a temporary error (which should be logged and retried).
+*/
```
clemens-msupply, line 6: You could move this comment as a doc comment to: `pub fn send_email(`

clemens-msupply (approved):
Looks good to me (untested, please let me know if you think I should setup the environment to do a full test!)

backend/service/src/notification/mod.rs
```diff
@@ -105,52 +107,143 @@ impl NotificationServiceTrait for NotificationService {
...
+                            } else {
+                                log::error!(
+                                    "Temporarily failed to send email {} to {} - {:?}",
+                                    notification.id,
+                                    notification.to_address,
+                                    send_error
+                                );
+                                notification.error_message = Some(format!("{:?}", send_error));
+                                notification.status = NotificationEventStatus::Errored;
+                                notification.retries += 1;
```
clemens-msupply, line 169: Shouldn't you always increase the retries count?

backend/service/src/notification/mod.rs
```diff
@@ -105,52 +107,143 @@ impl NotificationServiceTrait for NotificationService {
...
+                                } else {
+                                    log::error!(
+                                    "Temporarily failed to send telegram message {} to {} - {:?}",
+                                    notification.id,
+                                    notification.to_address,
+                                    e
+                                );
+                                    notification.error_message = Some(format!("{:?}", e));
+                                    notification.status = NotificationEventStatus::Errored;
+                                    notification.retries += 1;
```
clemens-msupply, line 228: Nit: move `notification.retries += 1;` out of the if conditions to make sure you always update it...

backend/coldchain/src/lib.rs
jmbrunskill: I think I like `NotificationTarget` will try it and see how it feels going forward.

backend/service/src/email/send.rs
jmbrunskill: Done

backend/service/src/notification/mod.rs
jmbrunskill: I've changed to this logic, as I think it's safer as you say less risk of missunderstanding in future.
I was trying to show the retries as 0 when it didn't re-try. I think this is fixed by recording send attempts, instead of re-tries...

Issue #116, opened 2023-09-08
How to Configure SQL params for cold chain 


jmbrunskill:
## Is your feature request related to a problem? Please describe

If a parameterised sql recipient query is selected in a configuration the user will need to specify which parameters they want to use with the query.

## Describe the solution you'd like

A unified parameter approach should be taken where if parameters are needed for some part of a configuration, the parameter configuration tool should pop up something like this.

![image](https://github.com/openmsupply/notify/assets/8287373/b87a85c4-1c11-4ce3-9d15-806bc10ee582)

Parameters should be deduped so they only appear once.

If a configuration needs to run with different sets of parameters, this should be possible as well (not an option when configuring a single query...)

## Describe alternatives you've considered

JSON config field

## Additional context

See also: #107 - don't use a modal!

PR #120, opened 2023-09-12
107 move notification configs from a modal to a page

### LINKED ISSUE #107: Move NotificationConfigs from a modal to a page...

Having things in a modal feels a bit wrong from a UX perspective, particularly when we have modals to select recipeinets etc.

Similar to recipient lists where you navigate to a page like http://localhost:3007/recipients/lists/302bb909-3098-4011-9754-32e844c7f224

Would allow us to use tabs if needed for different configuration portions if needed e.g. selecting SQL queries in a tab?

jmbrunskill:
Closes #107 

# 👩🏻‍💻 What does this PR do?
 

Moves to using a actual page rather than modal to manage notification configurations, which should give more flexibility and allow PRs such as #116 to add more features such as parameters.

The ui is modeled of open mSupply with a app bar and app footer design.

image

Right now only the cold chain config is in this branch. Similar config will need to be updated for the scheduled reports and any other future configuration types.

I'm not sure if the generics are actually required here, but tried not to change things too much from what Lache had done with modals.


frontend/packages/system/src/Notifications/ListView/ListView.tsx
```diff
@@ -13,16 +13,36 @@ import {
   createTableStore,
   useColumns,
 } from '@common/ui';
-import { NotificationsModal } from '../Modals/NotificationsModal';
 import { useEditModal, useQueryParamsState } from '@common/hooks';
 import { NotificationConfigRowFragment, useNotificationConfigs } from '../api';
 import { useDeleteNotificationConfig } from '../api/hooks/useDeleteNotificationConfig';
+import { ConfigKind, useNavigate } from 'packages/common/src';
```
mark-prins, line 19: importing directly from `packages/` like this is fine in the notify app - but breaks tests in open mSupply 🤷 
It makes this suggestion optional - though the import is neater looking if you use the package name:
```suggestion
import { ConfigKind, useNavigate } from '@notify-frontend/common';
```

frontend/packages/system/src/Notifications/ListView/ListView.tsx
```diff
@@ -13,16 +13,36 @@ import {
...
+import { configRoute } from '../navigate';
+import { CreateConfigKindModal } from '../Pages/CreateConfigKindModal';
```
mark-prins, line 21: I found the name misleading - the button is for a `New Notification` : the modal sounds like it's creating a `ConfigKind`. Perhaps `NewNotificationModal` or `CreateNotificationModal` 🤷

frontend/packages/system/src/Notifications/Pages/Base/BaseNotificationEditPage.tsx
```diff
@@ -0,0 +1,122 @@
...
+                height={64}
+              >
+                <Box flex={1} display="flex" justifyContent="flex-end" gap={2}>
+                  <ButtonWithIcon
+                    shrinkThreshold="lg"
+                    Icon={<CloseIcon />}
+                    label={t('button.close')}
+                    color="secondary"
+                    sx={{ fontSize: '12px' }}
+                    onClick={() => navigateUpOne()}
```
mark-prins, line 100: no need to create a new function here
```suggestion
                    onClick={navigateUpOne}
```

frontend/packages/system/src/Notifications/Pages/CreateConfigKindModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+interface SelectNotificationTypeModalProps {
+  isOpen: boolean;
+  onClose: () => void;
+}
+
+export const CreateConfigKindModal = ({
+  isOpen,
+  onClose,
+}: SelectNotificationTypeModalProps) => {
+  const t = useTranslation(['system']);
```
mark-prins, line 27:
```suggestion
  const t = useTranslation('system');
```

frontend/packages/system/src/Notifications/Pages/CreateConfigKindModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+                    title: title,
+                  },
+                }).then(() => {
+                  navigate(configRoute(kind, id));
+                });
+              }
+            }}
+            isLoading={isLoading}
+            startIcon={<ArrowRightIcon />}
+            variant="contained"
```
mark-prins, line 61: not required - the BaseButton is also setting the variant to `contained`

mark-prins (approved):
👍 Very nicely done!

I have a few layout suggestions, which I've added to a [new branch](https://github.com/openmsupply/notify/compare/107-move-notificationconfigs-from-a-modal-to-a-page...107-move-notification-configs-suggestion) as they touch a few files.

I've added some spacing to this modal:

Screenshot 2023-09-15 at 12 44 08 PM

like this:
image

And thought that the recipient list would be better with a label:

Screenshot 2023-09-15 at 1 36 19 PM

there's a few other comments, all minor bits tho

frontend/packages/system/src/Notifications/ListView/ListView.tsx
jmbrunskill: I rely too much on the auto importer :(
Thanks for picking this up!

PR #128, opened 2023-09-15
#102 disable notification configs

### LINKED ISSUE #102: Disable Notification Configs

It should be possible to disable a notification if the user wants to stop it from being generated without completely deleting it…

Notification Config should have a mechanism to disable them.

Configurations could have a disable/enable button on them like this:

image
image

Or it could be setup at the list view...

image

Either way, we'd need a similar thing in the backend.

The notification config row record will need a new field `disabled` which can be used to turn off a configuration if the user doesn't want it to be running for a while.

https://github.com/openmsupply/notify/blob/054c4dc5d8bd9c61e17f4fb6af09b1b86136bec9/backend/repository/src/db_diesel/notification_config_row.rs#L9

This new field would need to be configured all the way through to the graphql layer.

Having a dedicated graphql request to enable/disable a config might be worthwhile, or we could just use the 
update_notification_config one.

We could create the field as `disabled_at` with a timestamp.
This would allow us to show in the ui when it was disabled which would be useful e.g. this was disable today, vs last week vis last year etc.
https://medium.com/@munza/why-when-should-you-use-timestamp-instead-of-boolean-as-database-field-type-bdc8f83a8c21

## Additional context

- It might be good to do issue #107 (change from using a modal for config) before this work?

Mai21:
Closes #102

# 👩🏻‍💻 What does this PR do?
 
- Added the checkbox to change the notification status enabled/disabled.
Screenshot 2023-09-15 at 4 26 32 PM

- Added status column in the Notifications list view
Screenshot 2023-09-15 at 4 26 50 PM

- [ ] Check/Uncheck the Enable checkbox and save
- [ ] In the list, check if Status is updated properly

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2


frontend/packages/system/src/Notifications/types.ts
```diff
@@ -1,12 +1,13 @@
...
 import { NotificationConfigRowFragment } from './api';
 
-type BaseConfig = Pick<NotificationConfigRowFragment, 'id' | 'kind' | 'title'>;
+type BaseConfig = Pick<NotificationConfigRowFragment, 'id' | 'kind' | 'title' | 'status'>;
 
 // TODO: this should go away once recipient/list ids come through from the backend on the base config
 export interface BaseNotificationConfig extends BaseConfig {
   recipientIds: string[];
   recipientListIds: string[];
+  status: ConfigStatus;
```
jmbrunskill, line 10: I don't think this line is needed as we already have it from `BaseConfig`

backend/repository/src/db_diesel/notification_config_row.rs
```diff
@@ -11,6 +11,7 @@ table! {
         title -> Text,
         kind -> crate::db_diesel::notification_config_row::NotificationConfigKindMapping,
         configuration_data -> Text,
+        status -> crate::db_diesel::notification_config_row::NotificationConfigStatusMapping,
```
jmbrunskill, line 14: If you don't set a default value for this field, we need to use 
`status -> Nullable<crate::db_diesel::notification_config_row::NotificationConfigStatusMapping>,`
instead.

backend/repository/src/db_diesel/notification_config_row.rs
```diff
@@ -50,6 +76,7 @@ pub struct NotificationConfigRow {
     // this is actually stringified JSON - would be better to store as JSON, however
     // it would appear the diesel JSON types are only available if the postgres feature is enabled...
     pub configuration_data: String,
+    pub status: NotificationConfigStatus,
```
jmbrunskill, line 79: If you don't set a default value for this field, we would need to use 
` pub status: Option<NotificationConfigStatus>,`
instead.

jmbrunskill (approved):
Great work Mai,

Just need to fix that not-null issue and it's ready to merge.

frontend/packages/common/src/hooks/useDialog/useDialog.tsx
```diff
@@ -29,6 +29,7 @@ export interface ModalProps {
     } & React.RefAttributes<unknown>
   >;
   okButton?: JSX.Element;
+  statusCheckbox?: JSX.Element;
```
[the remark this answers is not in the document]
Mai21: I got it, I'll leave it as it is for now!

frontend/packages/system/src/Notifications/types.ts
Mai21: Fixed it!

backend/repository/migrations/2023-09-12-012051_add_status_to_notification_config/up.sql
```diff
@@ -0,0 +1,2 @@
+-- Your SQL goes here
+ALTER TABLE notification_config ADD COLUMN status TEXT;
```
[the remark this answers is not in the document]
Mai21: I agree with this, fixed it.

jmbrunskill:
@Mai21  Can you review again now that I've merge my other branch?

Looks like this...
image

jmbrunskill (approved):
Looks good to me, will double check with Mai before merging

Mai21:
@jmbrunskill Toggle Switch is a good idea:) Looks good to me!

Issue #112, opened 2023-09-04
Cold chain: Rename cold chain alerts in configuration modal


adamdewey:
## Is your feature request related to a problem? Please describe

Update language to match wording used elsewhere in cold chain solution. 

If possible a mouse-over would be good to de-clutter the screen too

## Describe the solution you'd like

See green highlights: 

<img width="1523" alt="image" src="https://github.com/openmsupply/notify/assets/19318937/c8c39940-5590-41bc-878d-f70762e69b8d">

Issue #111, opened 2023-09-04
Cold chain: Add section titles in configuration modal


adamdewey:
## Is your feature request related to a problem? Please describe

It would be good to add titles to sections of the configuration modal to make it easier for the user to understand.

## Describe the solution you'd like

See green highlights: 

<img width="1523" alt="image" src="https://github.com/openmsupply/notify/assets/19318937/1e22103f-06c7-4f45-91c8-47ba536233a8">

adamdewey:
Hey @jmbrunskill and @Mai21 - I see #131 says that is closes this issue, but actually there were a few more rearrangements and section headers (highlighted in green in the screenshot)

Would it be possible to do those too before closing this issue?

Cheers!

Mai21:
@adamdewey We changed Set up CC UI from the modal to the page. And recipients are now under the title.
Do we still need to add the section headers for recipients?
I'll add Preference header and send confirmation message when alert is resolved, fixed the title "Select Locations" to "Stoes and Locations".
![Screenshot 2023-10-02 at 11 49 05 AM](https://github.com/openmsupply/notify/assets/36495540/1740b72f-48b6-4937-99bd-f418d042e2af)

adamdewey:
@Mai21 - nice!

I think it would be good to add the recipients title too, just so it's obvious it's another configuration section - it can stay at the top though, all good!

PR #131, opened 2023-09-19
#110 cold chain 'no temperature data' alert

### LINKED ISSUE #110: Cold chain: 'No temperature data' alert

In current Grafana dashboards the user is sent an alert if the server does not receive temperature data for a certain period of time. 

This allows the local user to investigate the cause and attempt to rectify (plug power back into tablet, restart WiFi etc). 

This feature request is to add the same to Notify

Additional alert added to cold chain alert configuration (see green highlights here): 

image

If no temperature logs are received by the server for X amount of time then the alert should trigger

Mai21:
Closes #110, #111, #112, 

# 👩🏻‍💻 What does this PR do?
 
- Added the section title
- Added "i" icon to show more information about temperature threshold
- Added the config "Send no temperature data alert after: " with time frame
- Added "i" icon to show more information about no temperature data alert
Screenshot 2023-09-22 at 2 36 06 PM

- [ ] When checking this config, user can input time and select time type (Seconds/Minutes/Hours)
- [ ] When unchecking this config, user can not change time
- [ ] After saving Notificatios, reopen it and check if the values are properly saved


jmbrunskill (approved):
Looks great Mai.

I commited a tiny tweak for the wording and removed the `:` as I think it looks better without it when we have the 'i' icon.

Mai21:
@jmbrunskill I added the section (#111) and 'i' about temperature thresholds (#112).

frontend/packages/common/src/intl/locales/en/system.json
```diff
@@ -23,7 +23,8 @@
   "label.coldchain-confirm-ok-alerts": "Send temperature OK confirmation",
   "label.coldchain-high-temp-alerts": "Send high temperature alerts (Limits are based on your mSupply configuration)",
   "label.coldchain-low-temp-alerts": "Send low temperature alerts (Limits are based on your mSupply configuration)",
```
jmbrunskill, line 25:
```suggestion
  "label.coldchain-high-temp-alerts": "Send high temperature alerts",
  "label.coldchain-low-temp-alerts": "Send low temperature alerts",
```
This was doubling up with the `i` information

frontend/packages/system/src/Notifications/Pages/ColdChain/CCNotificationEditForm.tsx
```diff
@@ -63,6 +82,53 @@ export const CCNotificationEditForm = ({
...
+            onChange={newValue => onUpdate({ noDataInterval: newValue })}
+            sx={{ width: '60px' }}
+          />
+
+          <Select
+            value={draft.noDataUnits}
+            disabled={!draft.noData}
+            onChange={e =>
+              onUpdate({
+                noDataUnits: e.target.value as CCNotification['noDataUnits'],
```
jmbrunskill, line 122: I'm keen to avoid using `as` where we can. We can possibly avoid it here if we use an enum rather than a type literal. e.g.   `noDataUnits: 'seconds' | 'minutes' | 'hours';`
Here's an example of what we could do. 

https://github.com/openmsupply/notify/compare/%23110-Cold-chain-'No-temperature-data'-alert...%23110-Cold-chain-enum-example

I think it's a little bit safer this way, and easier to update too. I think we should probably remove the `seconds` option at it might create too many notifications, especially if someone says remind me every 5 seconds...

jmbrunskill (approved):
Looks good, a couple of suggestions provided.
When you're ready go ahead an merge.

frontend/packages/common/src/intl/locales/en/system.json
Mai21: I forgot to fix them!

frontend/packages/system/src/Notifications/Pages/ColdChain/CCNotificationEditForm.tsx
Mai21: @jmbrunskill I see! I will merge your example branch and update to remove `seconds`!

Issue #140, opened 2023-09-21
Refactor Name Validation code and tests


jmbrunskill:
## Is your feature request related to a problem? Please describe

Partially due to a lot of copy and pasting, we've ended up with a lot of validation functions like...

`check_list_name_is_appropriate_length`

`check_list_name_doesnt_contain_special_characters`

## Describe the solution you'd like

Shared functions should be moved into appropriate places and code updated.
For example some of the code has been refactored to sit in `util/src/validate.rs`

`is_valid_name` for example checks for invalid chars.
but the test cases for this are copied and pasted into lots of different services.

## Describe alternatives you've considered

Leave it how it is...

## Additional context

Leaving this as a separate issue to not distract from main task...

PR #141, opened 2023-09-21
124 sql query mangement (Backend) aka boilerplate

### LINKED ISSUE #124: SQL Query Management Backend

To configure Scheduled Reports users need to be able to manage SQL Queries that become parameters/variables in their templates

UI will be done in. https://github.com/openmsupply/notify/issues/145

Something similar to SQL Recipients, backend to store the queries and graphql endpoints to update/save them.
New table to store the queries

SQL queries as templates on disk

jmbrunskill:
Closes #124 - Basic Backend Only

Next up is https://github.com/openmsupply/notify/issues/145 (Frontend)

# 👩🏻‍💻 What does this PR do?
 

 (these queries will be used with in other Notification Configs, primarily `Scheduled` Notifications)
- [ ] Add a repository for `notification_query`
- [ ] Add a service `notification_query_service`
- [ ] Add a graphql crate `graphql_notifiation_query`

In #145 PR, We may need to make backend changes that are more specific to this use case (running a query and handling the responses for example.)

Note: also created a refactor issue https://github.com/openmsupply/notify/issues/140 given we're accumulating a lot of copy and pasted code validation code...


backend/graphql/notification_query/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+use repository::NotificationQueryFilter;
+use repository::PaginationOption;
+use service::auth::{Resource, ResourceAccessRequest};
+
+#[derive(Default, Clone)]
+pub struct NotificationQueryQueries;
+
+#[Object]
+impl NotificationQueryQueries {
+    pub async fn notification_querys(
```
jmbrunskill, line 21: Should this be `notification_queries`?
I quite like that a search for `notification_query` would find this, but also feels kind of wrong...

backend/util/src/validate.rs
```diff
@@ -1,7 +1,7 @@
 use regex::Regex;
 
 lazy_static! {
-    static ref SPECIAL_CHARS_RE: Regex = Regex::new(r"[^ 0-9A-Za-z_\-@.+:/()&]").unwrap();
+    static ref SPECIAL_CHARS_RE: Regex = Regex::new(r"[^ 0-9A-Za-z_\-@.+:/(),&]").unwrap();
```
jmbrunskill, line 4: I want to be able to create a name like `mSupply users, with region and country`

backend/graphql/notification_query/src/mutations/update.rs
```diff
@@ -0,0 +1,63 @@
...
+use crate::types::NotificationQueryNode;
+
+use super::{map_error, ModifyNotificationQueryResponse};
+#[derive(InputObject, Clone)]
+pub struct UpdateNotificationQueryInput {
+    pub id: String,
+    pub name: Option<String>,
+    pub description: Option<String>,
+    pub query: Option<String>,
+    pub required_parameters: Option<Vec<String>>,
```
roxy-dao, line 18: Does this need to be an Option? Can't just check if vec is empty?

backend/graphql/notification_query/src/types/notification_query.rs
```diff
@@ -0,0 +1,86 @@
+use async_graphql::{Object, SimpleObject, Union};
+use graphql_core::simple_generic_errors::NodeError;
+
+use repository::NotificationQuery;
+use service::ListResult;
+use util::usize_to_u32;
+
+#[derive(Union)]
+pub enum NotificationQuerysResponse {
```
roxy-dao, line 9: NotificationQueries 😉

backend/repository/src/db_diesel/notification_query_row.rs
```diff
@@ -0,0 +1,90 @@
...
+        id TEXT NOT NULL PRIMARY KEY,
+        name TEXT NOT NULL,
+        description TEXT NOT NULL,
+        query TEXT NOT NULL,
+        required_parameters TEXT NOT NULL, -- JSON e.g. {"limit": "number"}
+        created_at TIMESTAMP NOT NULL,
+        updated_at TIMESTAMP NOT NULL
+    );
+
+*/
```
roxy-dao, line 20: delete?

backend/service/src/notification_query/tests/create.rs
```diff
@@ -0,0 +1,149 @@
...
+        let result = service.create_notification_query(
+            &context,
+            CreateNotificationQuery {
+                id: new_notification_query_id.clone(),
+                name: "new_notification_query".to_string(),
+                ..Default::default()
+            },
+        );
+
+        if !result.is_ok() {
```
roxy-dao, line 136: Do you need to check this in create success test? 🤔

backend/service/src/notification_query/tests/update.rs
```diff
@@ -0,0 +1,184 @@
...
+                    id: "id1".to_string(),
+                    name: Some("name_for_id1".to_string()),
+                    description: None,
+                    ..Default::default()
+                },
+            )
+            .unwrap();
+
+        // updated
+        assert_eq!(updated_notification_query.name, "name_for_id1".to_string());
```
roxy-dao, line 160: There's an inline_edit function in om that allows you to edit the fields that have been changed for original row and check against updated row

backend/service/src/notification_query/validate.rs
```diff
@@ -0,0 +1,91 @@
...
+                )?;
+
+            Ok(notification_querys.is_empty())
+        }
+    }
+}
+
+// TODO: Refactor as part of https://github.com/openmsupply/notify/issues/140
+pub fn check_list_name_is_appropriate_length(name: &str) -> Result<bool, RepositoryError> {
+    Ok(name.trim().len() >= 3 && name.len() <= 70)
```
roxy-dao, line 51:
```suggestion
    Ok(name.trim().len() >= 3 && name.trim().len() <= 70)
```?

File: backend/util/src/validate.rs (hunk 2 above)
Comment (roxy-dao) [reply backend/util/src/validate.rs:4]: Tried using regex checker for this... but it's not giving me any matches 😭

**Review by roxy-dao — CHANGES_REQUESTED** [review]
Looks good James! Just a few comments (:

File: backend/graphql/notification_query/Cargo.toml [hunk 9]
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "graphql_notification_query"
+version = "0.1.0"
+edition = "2018"
```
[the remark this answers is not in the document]
jmbrunskill: If I set it to 2021 (this is the new default) I get this message

```
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
```

Haven't looked into it more deeply yet maybe `workspace.resolver = "2"` is somehow awesome for us?

backend/graphql/notification_query/src/mutations/update.rs
jmbrunskill: If we set it to `Option::None`, it means diesel will not update that field.
Empty Vec means update the record and set it to an empty array.

backend/graphql/notification_query/src/mutations/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+pub enum ModifyNotificationQueryResponse {
+    Response(NotificationQueryNode),
+}
+
+pub fn map_error(error: ModifyNotificationQueryError) -> Result<ModifyNotificationQueryResponse> {
+    let formatted_error = format!("{:#?}", error);
+
+    let graphql_error = match error {
+        // Standard Graphql Errors
+        ModifyNotificationQueryError::NotificationQueryDoesNotExist => {
```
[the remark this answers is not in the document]
jmbrunskill: Fair enough, I think I find it easier to see and understand on different lines, although I do like it for these very related errors though.

backend/repository/src/db_diesel/notification_query_row.rs
jmbrunskill: Yeah I guess so, will get out of date. I find it a useful reference though...
Hmmm

backend/service/src/notification_query/tests/create.rs
jmbrunskill: I'm not sure about this pattern, but I've found it kind of helpful when debugging tests. E.g. In this test we're asserting that this result is ok. If it's not ok, we want to see what the actual error was. Probably a better idea is to do assert_eq!() then you'll see the error if it's not ok... Will try that...

backend/util/src/validate.rs
jmbrunskill: image

backend/service/src/notification_query/tests/create.rs
jmbrunskill: Ok, that's not easy either, as we need to recreate the whole row to check it's ok. Changed to unwrap...

backend/service/src/notification_query/validate.rs
jmbrunskill: I guess it could be this, but the point is really to make sure they have a put something other than spaces in the field in, and b) haven't made it too big. (this is copy and pasted from Lache's code)

backend/graphql/notification_query/src/lib.rs
jmbrunskill: Cool thanks for the feedback, updated https://github.com/openmsupply/notify/pull/141/commits/ee6986089be9106897a9895831a966e2967be4d8

backend/util/src/validate.rs
jmbrunskill: Confusion might be that it's matching any non-standard characters.
The list of allowed chars is getting pretty long though, I wonder if it's worth having this check at all? Just allow what ever?

roxy-dao: Think it's fine to have the check

roxy-dao (approved):
LGTM~

Issue #138, opened 2023-09-20
Schduled Notifications should send to configured recipients


jmbrunskill:
## Is your feature request related to a problem? Please describe

Currently we send to a test recipient. We should look up the correct recipients to send too, including sql recipients?

PR #152, opened 2023-09-26
Parse recipient_ids & lists into their own fields

### LINKED ISSUE #148: Store recipient information separately to other notification config

Store recipient information separately to other notification config

When processing a notification, we need to make sure that shared information (such as recipients) are made available to the backend. Treating these fields separately makes this easier to deal with.

jmbrunskill:
Closes #148 

# 👩🏻‍💻 What does this PR do?
 

Move recipient configs into their own fields in the database, making it easier to shared backend code to extract the recipients.

Needed for #138

### DIFF
--- a/backend/repository/src/db_diesel/notification_config.rs
+++ b/backend/repository/src/db_diesel/notification_config.rs
@@ -7,12 +7,11 @@ use super::{
 use crate::{
     diesel_macros::{apply_equal_filter, apply_sort_no_case, apply_string_filter},
     repository_error::RepositoryError,
-    EqualFilter, NotificationConfigKind, NotificationConfigRow, Pagination, Sort, StringFilter, NotificationConfigStatus, 
+    EqualFilter, NotificationConfigKind, NotificationConfigRow, NotificationConfigStatus,
+    Pagination, Sort, StringFilter,
 };
 
 use diesel::{dsl::IntoBoxed, prelude::*};
-
-pub type NotificationConfig = NotificationConfigRow;
 
 #[derive(Clone, Default, Debug, PartialEq)]
 pub struct NotificationConfigFilter {
@@ -48,14 +47,14 @@ impl<'a> NotificationConfigRepository<'a> {
     pub fn query_by_filter(
         &self,
         filter: NotificationConfigFilter,
-    ) -> Result<Vec<NotificationConfig>, RepositoryError> {
+    ) -> Result<Vec<NotificationConfigRow>, RepositoryError> {
         self.query(Pagination::new(), Some(filter), None)
     }
 
     pub fn query_one(
         &self,
         filter: NotificationConfigFilter,
-    ) -> Result<Option<NotificationConfig>, RepositoryError> {
+    ) -> Result<Option<NotificationConfigRow>, RepositoryError> {
         Ok(self.query_by_filter(filter)?.pop())
     }
 
@@ -64,7 +63,7 @@ impl<'a> NotificationConfigRepository<'a> {
         pagination: Pagination,
         filter: Option<NotificationConfigFilter>,
         sort: Option<NotificationConfigSort>,
-    ) -> Result<Vec<NotificationConfig>, RepositoryError> {
+    ) -> Result<Vec<NotificationConfigRow>, RepositoryError> {
         let mut query = create_filtered_query(filter);
 
         if let Some(sort) = sort {
@@ -81,7 +80,7 @@ impl<'a> NotificationConfigRepository<'a> {
             .offset(pagination.offset as i64)
             .limit(pagination.limit as i64);
 
-        let result = final_query.load::<NotificationConfig>(&self.connection.connection)?;
+        let result = final_query.load::<NotificationConfigRow>(&self.connection.connection)?;
         Ok(result)
     }
 }
--- a/backend/repository/src/db_diesel/notification_config_row.rs
+++ b/backend/repository/src/db_diesel/notification_config_row.rs
@@ -151,10 +151,10 @@ impl<'a> NotificationConfigRowRepository<'a> {
         Ok(result)
     }
 
-    pub fn set_next_due_by_id(
+    pub fn set_last_run_by_id(
         &self,
         id: &str,
-        last_run: Option<NaiveDateTime>,
+        last_run: NaiveDateTime,
         next_due: Option<NaiveDateTime>,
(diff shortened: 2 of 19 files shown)


backend/repository/src/db_diesel/notification_config.rs
```diff
@@ -48,14 +47,14 @@ impl<'a> NotificationConfigRepository<'a> {
     pub fn query_by_filter(
         &self,
         filter: NotificationConfigFilter,
-    ) -> Result<Vec<NotificationConfig>, RepositoryError> {
+    ) -> Result<Vec<NotificationConfigRow>, RepositoryError> {
```
andreievg, line 50: Was wondering why remove the type ?(extra PR changes, and the type alias pattern makes it easier to extned if it becomes a composite type)

backend/repository/src/db_diesel/notification_config_row.rs
```diff
@@ -144,4 +150,21 @@ impl<'a> NotificationConfigRowRepository<'a> {
             .load::<NotificationConfigRow>(&self.connection.connection)?;
         Ok(result)
     }
+
+    pub fn set_next_due_by_id(
+        &self,
+        id: &str,
+        last_run: Option<NaiveDateTime>,
```
andreievg, line 157: Both can be non optional, or do we set them to null sometimes ?

backend/service/src/notification_config/query.rs
```diff
@@ -10,11 +11,43 @@ use crate::{
...
+impl From<NotificationConfigRow> for NotificationConfig {
+    fn from(row: NotificationConfigRow) -> Self {
+        NotificationConfig {
+            id: row.id,
+            title: row.title,
+            kind: row.kind,
+            configuration_data: row.configuration_data,
+            status: row.status,
+            parameters: row.parameters,
+            recipient_ids: serde_json::from_str(&row.recipient_ids).unwrap_or_default(),
```
andreievg, line 41: I think instead of a new service type, a method on NotificationConfig (turn into struct). Also destrcuture structure back help comiple aid in future field additions

frontend/packages/system/src/Notifications/Pages/ColdChain/parseConfig.ts
```diff
@@ -36,10 +36,10 @@ export function parseColdChainNotificationConfig(
       reminderInterval,
       reminderUnits,
       locationIds,
-      recipientIds,
-      recipientListIds,
+      recipientIds: recipientIds ?? config.recipientIds,
```
andreievg, line 39: I don't quite understand why we would have these field in configurationData and in config.recepientIds ? (is that for backward compatibility, i thought we havne't deployed anywhere yet ?)

frontend/packages/system/src/Notifications/Pages/Scheduled/parseConfig.ts
```diff
@@ -19,6 +19,10 @@ export function parseScheduledNotificationConfig(
       id: config.id,
       title: config.title,
       kind: config.kind,
+      parameters: config.parameters,
```
andreievg, line 22: maybe best to destructure and restructure or spread ? (easy to miss this ?)

andreievg (approved):
Looks good thanks, some minor suggestions and questions

andreievg (approved):
Looks good thanks, some minor suggestions and questions

backend/repository/src/db_diesel/notification_config.rs
jmbrunskill: I was trying to remember why but pretty sure it's because of the service type `NotificationConfig` didn't want duplication in the name. I see you have comments about that type, so can address it with you there...

backend/repository/src/db_diesel/notification_config_row.rs
jmbrunskill: In cold chain, I'm currently just setting the `last_run` time, not the `next_due`.
That being said I'm open to changing it...

jmbrunskill: The re-introduction of this function (rather than Config Row update was the change to the service type) if we get rid of that we can go back to updating the whole row.
I'm a bit of a fan of these more specific methods though for some reason, maybe just previous expereince/approaches in other code bases...

backend/service/src/notification_config/query.rs
jmbrunskill: I'm not sure I fully understand your idea here.
You thinking methods like `get_recipient_ids`?

That does sound like quite a nice approach...

frontend/packages/system/src/Notifications/Pages/ColdChain/parseConfig.ts
jmbrunskill: Yup, backwards compat, thought it didn't hurt but yeah I guess it is kind of confusing....
An we do have something deployed, which now that I think about it might need this...

frontend/packages/system/src/Notifications/Pages/Scheduled/parseConfig.ts
jmbrunskill:
```
    return {
      ...defaultSchedulerNotification,
      ...config,
      ...JSON.parse(config.configurationData),
    };
```
Like this?
Seems like a good idea!

Not clear on how the destructing would look, maybe that's better as the developer sees what's changed where?

frontend/packages/system/src/Notifications/Pages/ColdChain/parseConfig.ts
andreievg: Maybe a comment please

backend/service/src/notification_config/query.rs
jmbrunskill: Changed to desctructure

frontend/packages/system/src/Notifications/Pages/ColdChain/parseConfig.ts
jmbrunskill: https://github.com/openmsupply/notify/pull/152/commits/c75d9e9df522be745fa95caba20f7d27f82876d8

backend/repository/src/db_diesel/notification_config.rs
jmbrunskill: As discussed leaving this for now (partly for time pressures, will look at  refactoring options in future)

Issue #118, opened 2023-09-08
Handle invalid telegram html entities better


jmbrunskill:
## Describe the bug

"missing field `result` at line 1 column 126-{\"ok\":false,\"error_code\":400,\"description\":\"Bad Request: can't parse entities: Unsupported start tag \\\"h3\\\" at byte offset 0\"}"

## To Reproduce

Steps to reproduce the behaviour:

1. Create a notification with an `h3` tag
2. Send notification - causes telegram error

## Expected behaviour

Not entirely sure?? Would be good to show what the problem is better...

jmbrunskill:
Should we just support markdown in notify?

https://crates.io/crates/pulldown-cmark

PR #156, opened 2023-09-28
Render templates

### LINKED ISSUE #136: Scheduled Notification Plugin should render the configured templates

Following on from #125 when processing scheduled notifications, the code should attempt to render the correct template with configured parameters.

jmbrunskill:
Closes #136 

# 👩🏻‍💻 What does this PR do?
 

This PR adds templating functionality to Scheduled Notifications.

With this PR notifiy should now send the notification of the configured template rather than a test notification.

It adds a mechanism to access the shared tera instance in the service provider (and therefore provides a mechanism to extend it)
Sending notifications can now be created using a tera template loaded from disk, or from one defined in the scheduled task.

I'd recommend testing it using mailhog and/or telegram (if you have a bot setup)
It's helpful to save the notification with a run time 1 minute from now, so there's time for it to be processed after you save it.

image

Here's an example notification in email:
image

and in telegram

image

image

At this stage you can use the recipient name as a parameter, you can reference it in a template like this...
{{ recipient.name }} (id & to_address are also available)

I'd recommend trying out different broken configurations to see what happens.

> Right now there's an issue with telegram alerts if you use unsupport HTML tags `<p></p>` for example!
- [ ] https://github.com/openmsupply/notify/issues/118

### DIFF
--- a/backend/service/src/notification/enqueue.rs
+++ b/backend/service/src/notification/enqueue.rs
@@ -4,6 +4,7 @@ use repository::{
     NotificationType, RecipientRow,
 };
 use serde::Serialize;
+use tera::{Context, Tera};
 use util::uuid::uuid;
 
 use crate::service_provider::ServiceContext;
@@ -30,9 +31,15 @@ impl From<RecipientRow> for NotificationTarget {
 }
 
 #[derive(Debug)]
+pub enum TemplateDefinition {
+    TemplateName(String),
+    Template(String),
+}
+
+#[derive(Debug)]
 pub struct NotificationContext {
-    pub title_template_name: Option<String>,
-    pub body_template_name: String,
+    pub title_template: Option<TemplateDefinition>,
+    pub body_template: TemplateDefinition,
     pub recipients: Vec<NotificationTarget>,
     pub template_data: serde_json::Value,
 }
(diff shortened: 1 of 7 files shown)


backend/service/src/notification/enqueue.rs
```diff
@@ -46,12 +53,57 @@ pub fn create_notification_events(
...
+                    NotificationServiceError::GenericError(format!(
+                        "Failed to add body template to tera instance: {:?}",
+                        e
+                    ))
+                })?;
+            "body_template".to_string()
+        }
+    };
+
+    let mut tera_context = Context::from_serialize(notification.template_data.clone())
```
andreievg, line 98: I think `from_value` is more direct way to do this ?

backend/service/src/notification/enqueue.rs
```diff
@@ -67,53 +119,32 @@ pub fn create_notification_events(
...
+        let notification_queue_row = match tera.render(&body_template_name, &tera_context) {
+            Ok(body) => NotificationEventRow {
+                message: body,
+                ..base_row_with_title
+            },
+            Err(e) => {
+                log::error!("Failed to render notification title template: {:?}", e);
+                NotificationEventRow {
+                    status: NotificationEventStatus::Errored,
+                    error_message: Some(format!("{:?}", e)),
```
andreievg, line 146: Maybe add error from base_row_with_title (in case there is also title error ?)

andreievg (approved):
Looks good to me, I haven't tested but had a thorough look at the code

andreievg (approved):
Looks good to me, I haven't tested but had a thorough look at the code

frontend/packages/common/src/intl/locales/en/host.json
```diff
@@ -47,5 +47,6 @@
   "saving": "Saving...",
   "set-up-account": "Set up your account",
   "username-reset.explanation": "Enter your new log in username",
-  "users": "Users"
-}
+  "users": "Users",
+  "queries": "Queries"
```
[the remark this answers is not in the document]
jmbrunskill: It's just in the Url, @mark-prins  asked me to add this, not sure how it ended up in this PR though...

image

jmbrunskill:
> Looks good to me, I haven't tested but had a thorough look at the code. Great to see it all coming together, exciting !

No worries, the QA team is itching to test, so I think we can leave that to them.

backend/service/src/notification/enqueue.rs
jmbrunskill: I think it's ok how it is, if the body template doesn't render that' the important thing to fix.
I'm also hoping we have time to add a test function, so hopefully people will have tested their configs before saving/enabling them.

PR #163, opened 2023-10-02
#115 Dedup Notification Recipients

### LINKED ISSUE #115: Dedup Notification Recipients

When processing a Notification, it's possible to get the same recipient multiple times.
For example, the recipient could be part of two lists, or the sql query could return the same user multiple times.
We should remove any duplicates before processing the notification list to reduce issues with users being sent too many notifications.

Deduplicate based on `to_address` & `notification_type`
This means uses might receive both telegram and email notifications but that's ok for now.

- Carefully construct your notifications to avoid duplication, which is prone to use error

## Additional context

In `service/src/notification/enqueue.rs` there's a function called `create_notification_events`.
https://github.com/openmsupply/notify/blob/358aee93ef7769233474f059122172970f07dcab/backend/service/src/notification/enqueue.rs#L40

This takes a `NotificationContext` which contains a list of `NotificationTarget` which is essentially the same as a recipient.

https://github.com/openmsupply/notify/blob/358aee93ef7769233474f059122172970f07dcab/backend/service/src/notification/enqueue.rs#L16

(We are using a different struct type, as it might be possible for plugins to generate their own recipients in future)

Because these `NotificationTarget` records might come from an sql query, recipient list, recipients etc. There might be more than one copy of the same to_address in the notification context.

`create_notification_events` should remove any duplicate values from `notification.recipients` before looping through them to make sure we don't send multiple copies of the same alert to the same person!

Implementation options...
- https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.unique_by
- Built in `dedup_by` method https://doc.rust-lang.org/std/vec/struct.Vec.html#method.dedup_by (but this requires the vec to be sorted first.)
- Or we could write a custom function?

Mai21:
Closes #115

# 👩🏻‍💻 What does this PR do?
 

Added the code to dedup recipients by to_address.

Do we need to check the notification type (Telegram/Email) as well? 

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2

### DIFF
--- a/backend/service/src/notification/enqueue.rs
+++ b/backend/service/src/notification/enqueue.rs
@@ -49,8 +49,6 @@ pub fn create_notification_events(
     recipients.sort_by(|a, b| a.to_address.cmp(&b.to_address));
     recipients.dedup_by(|a, b| (a.to_address == b.to_address));
 
-    // TODO: Should this function use a notification config to get the template, users, etc?
-
     // Loop through recipients and create a notification for each
     for recipient in recipients {
         let notification_type = recipient.notification_type.clone();
@@ -173,7 +171,12 @@ mod test {
                     name: "test".to_string(),
                     to_address: "test@example.com".to_string(),
                     notification_type: NotificationType::Email,
-                }],
+                }, NotificationTarget {
+                    name: "test2".to_string(),
+                    to_address: "test@example.com".to_string(),
+                    notification_type: NotificationType::Email,
+                },
+                ],
                 template_data: serde_json::json!({}),
             },
         );
@@ -215,6 +218,10 @@ mod test {
                 body_template_name: "test_message/telegram.html".to_string(),
                 recipients: vec![NotificationTarget {
                     name: "telegram".to_string(),
+                    to_address: "-12345".to_string(),
+                    notification_type: NotificationType::Telegram,
+                }, NotificationTarget {
+                    name: "telegram2".to_string(),
                     to_address: "-12345".to_string(),
                     notification_type: NotificationType::Telegram,
                 }],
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
Looks good Mai.

Can you also update or add a test cases to check that it's working?

The easiest way to do this is to add a duplicate recipient to the `NotificationContext` for one of the existing tests like this...
```
                    NotificationTarget {
                        name: "test".to_string(),
                        to_address: "test@example.com".to_string(),
                        notification_type: NotificationType::Email,
                    },
                    NotificationTarget {
                        name: "duplicate_test".to_string(),
                        to_address: "test@example.com".to_string(),
                        notification_type: NotificationType::Email,
                    },
```

I did this to verify the logic was actually working :)

Having a test case, helps reviewers be sure the logic is correct, and helps us make sure we don't accidentally remove logic like this in future without realising it.

backend/service/src/notification/enqueue.rs
```diff
@@ -44,10 +44,15 @@ pub fn create_notification_events(
 ) -> Result<(), NotificationServiceError> {
     let repo = NotificationEventRowRepository::new(&ctx.connection);
 
+    // Dedup recipients by to_address
+    let mut recipients = notification.recipients.clone();
+    recipients.sort_by(|a, b| a.to_address.cmp(&b.to_address));
+    recipients.dedup_by(|a, b| (a.to_address == b.to_address));
+
     // TODO: Should this function use a notification config to get the template, users, etc?
```
jmbrunskill, line 52: Can you remove this todo now please?

Mai21:
@jmbrunskill Fixed in [3dc84dc](https://github.com/openmsupply/notify/pull/163/commits/3dc84dc59c5b1ee466c3fc3b437c531441ad040c)

jmbrunskill:
Looks good Mai, read to merge!

jmbrunskill (approved):
Looks good, let's merge

PR #164, opened 2023-10-02
161 windows service

### LINKED ISSUE #161: Windows Service

Enable notifiy to run as a windows service similar to open-mSupply

Non sucky service manager

jmbrunskill:
Closes #161 

# 👩🏻‍💻 What does this PR do?
 

Allows building the server as a windows service.

Register the service with a command like this:
`sc create Notify binpath="C:\Program Files\notify\notify_service.exe"`

Pretty much a copy and past from open-msupply, but a few false starts with small differences like logging config,

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -1050,16 +1050,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ctor"
-version = "0.1.21"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ccc0a48a9b826acdf4028595adc9db92caea352f7af011a3034acd172a52a0aa"
-dependencies = [
- "quote",
- "syn 1.0.109",
-]
-
-[[package]]
 name = "cxx"
 version = "1.0.81"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


mark-prins (approved):
🎉  LGTM!

It compiles and seems to run on windows.. though I couldn't find any file based logging, did I miss that? and I tried browsing to the site, but got nothing - are there some steps I'm missing here? Perhaps some more comprehensive testing notes would help.

The service has the problem which I think oms still has - if you mess things up ( like having no configuration directory) and start the service, then it won't start and won't stop. If rust panics while starting the service then it goes into an indeterminate state.

I had added the windows event logging to at least give some info when that happens.. and figured it was a one-off config / setup issue, so didn't go too far down the rabbit hole. I think it's ok as it is!

File based logging though.. that's very helpful when running as a service with no console to see messages in 😁

jmbrunskill:
>The service has the problem which I think oms still has - if you mess things up ( like having no configuration directory) and start the service, then it won't start and won't stop. If rust panics while starting the service then it goes into an indeterminate state.

Hmm I had it refuse to start but not get into a unable to stop or start state but haven't really done much testing. I'll ask the testers to look into it. I figure if we have an installer, we don't have the missing config directory for example. 

>File based logging though.. that's very helpful when running as a service with no console to see messages in 😁

Great point. I've added a new issue for this : https://github.com/openmsupply/notify/issues/173

PR #169, opened 2023-10-03
165 cold chain notification configuration should show the real sensors

### LINKED ISSUE #165: Cold Chain Notification Configuration should show the 'real' locations/stores

Right now a hard coded list of locations with sensors is shown in the configuration.

The mSupply Locations containing sensors should be shown and be available to configure.

## Acceptance Criteria

- [ ] The list of all active sensors configured in mSupply is displayed
- [ ] Saving the list of sensor works (able to save the list and reopen with the same list showing up)
- [ ] If sensor query fails (e.g. not connected to dashboard DB) an error should show
- [ ] Without a location, the notification shouldn't be able to be saved as `enabled`

jmbrunskill:
Closes #165 

# 👩🏻‍💻 What does this PR do?
 

Kapture 2023-10-03 at 17 57 39

This needs an mSupply postgres (AKA Dashboard) database setup.
Ideally with coldchain sensors configured, although might be interesting to figure how/what we should display if there are no sensor...

I'm not sure if the SQL Query in the front is a good, longer term plan.
I think the longer term plan (plugin queries perhaps?) should go in a seperate issue with design, and this might need to be merged for expediency, unless another quick to implement option is thought of.


backend/coldchain/src/alerts.rs
```diff
@@ -0,0 +1,147 @@
+use chrono::NaiveDateTime;
```
jmbrunskill, line 1: This file is really just a move from `lib.rs`

backend/coldchain/src/lib.rs
```diff
@@ -1,148 +1,61 @@
...
     service_provider::ServiceContext,
 };
 
-/*
+pub mod alerts;
 
-Temperature Alerts will look something like this...
------------------------
-High temperature alert!
+pub struct ColdChainPlugin {}
```
jmbrunskill, line 8: Making cold chain a plugin isn't really required for this PR, as it's currently using an SQL query in the frontend to retrieve data. I think it should somehow query the plugin instead, but leaving that for a separate PR & Issue.
Figured it doesn't hurt to keep this change in this PR.

frontend/packages/system/src/Notifications/Pages/ColdChain/SensorSelectionModal.tsx
```diff
@@ -0,0 +1,154 @@
...
+  }, [sensors]);
+
+  const onChangeSelectedQueries = (ids: string[]) => {
+    setSelectedIds(ids);
+  };
+
+  const onSubmit = async () => {
+    if (selectedIds.length === 0) {
+      setErrorMessage(t('messages.nothing-selected'));
+      //   return;
```
jmbrunskill, line 59: I'm not sure what we want here. Should someone be able to select no sensors, I can see why it could be useful when your playing around, but also see the arguement for requiring one...

frontend/packages/system/src/Notifications/Pages/ColdChain/SensorSelector.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const { isOpen, onClose, onOpen } = useEditModal();
+
+  const columns = useColumns<SensorData>([
+    {
+      key: 'name',
+      label: 'label.name',
+      width: 150,
+      sortable: true,
+      accessor: ({ rowData }) => {
+        return sensorDisplayName(rowData);
```
jmbrunskill, line 39: Should we render the store, location and sensor separately rather than display name?

frontend/packages/system/src/Notifications/Pages/ColdChain/SensorSelector.tsx
```diff
@@ -0,0 +1,73 @@
...
+  return (
+    <>
+      <SensorSelectionModal
+        sensors={records ?? []}
+        initialSelectedIds={selectedIds}
+        isOpen={isOpen}
+        onClose={onClose}
+        setSelection={setSelection}
+      />
+      <IconButton
```
jmbrunskill, line 57: Don't love this edit button, but keen for ideas how we could do it better, perhaps some examples from another app?

frontend/packages/system/src/Notifications/api/hooks/useColdChainSensors.ts
```diff
@@ -0,0 +1,37 @@
...
+  const sdk = getSdk(client);
+
+  const cacheKeys = ['COLDCHAIN_SENSORS'];
+
+  return useQuery(cacheKeys, async () => {
+    const sensorQuery =
+      "SELECT sn.id as id, s.name as store_name,coalesce(l.description, '') as location_name, sn.name as sensor_name FROM SENSOR sn JOIN store s ON sn.storeid = s.id LEFT JOIN location l on sn.locationid = l.id WHERE sn.is_active = true ORDER BY 2,3,4 LIMIT 1000";
+    const response = await sdk.getColdChainSensors({
+      sqlQuery: sensorQuery,
+      params: '{}',
```
jmbrunskill, line 29: Figured if we stuck with this approach, we could pass parameters to say search if we end up with too many sensors in a bit installation...

jmbrunskill:
Sorry I merged the windows service PR here, but shouldn't have. Please ignore commit [9d11fc3](https://github.com/openmsupply/notify/pull/169/commits/9d11fc3a193e2d992c7b506396f1b5b674175c44)

frontend/packages/system/src/Notifications/api/operations.graphql
```diff
@@ -48,3 +48,7 @@ mutation deleteNotificationConfig($id: String!) {
     }
   }
 }
+
+query getColdChainSensors($sqlQuery: String, $params: String) {
+  runSqlQueryWithParameters(sqlQuery: $sqlQuery, parameters: $params)
+}
```
clemens-msupply, line 54: oh didn't know that you don't need to select what you want to return does this just mean you return all fields? What about recursive queries?

clemens-msupply: Nit: just name it `runSqlQueryWithParameters`? Got confused that there actually is a query for cold chain sensors...

frontend/packages/system/src/Notifications/Pages/ColdChain/SensorSelector.tsx
clemens-msupply: yes I would say so...

frontend/packages/system/src/Notifications/Pages/ColdChain/CCNotificationEditForm.tsx
```diff
@@ -206,44 +191,18 @@ export const CCNotificationEditForm = ({
...
-                  {location.name}
-                </label>
-              </Grid>
-            );
-          })}
+          <SensorSelector
+            records={sensors ?? []}
+            selectedIds={draft.sensorIds}
+            setSelection={props => {
+              console.log('props', props);
```
clemens-msupply, line 201: debug output...

clemens-msupply (approved):
Regarding, your query question. As it looks like it the backend is kept minimalistic and just provides the core functionality(?). I think then it makes sense to move the logic to the frontend (into frontend plugins I guess).

Do we need backend plugins at all? If you do, it would make more sense to provide the sensor list through the backend though(?) But what would be the generic interface to communicate between frontend and backend plugins? the frontend plugin sends a json blob to its backend plugin?

Or you make cold chain sensor a build in feature and provide proper graphql queries but as you said this would be extra work.

One thing I noticed: `send_high_temperature_alert_telegram` could be moved into the test, right?

frontend/packages/common/src/intl/locales/en/system.json
```diff
@@ -86,5 +86,6 @@
   "helper-text.sql-query": "Your sql query can contain parameters, which are created using double curly braces within the query. For example: SELECT * FROM my_table WHERE id = {{param1}} AND name = {{param2}}",
   "label.test-sql-query": "Test SQL Query",
   "tooltip.manage-recipient-list": "Manage Recipient List",
-  "message.no-parameters": "No parameters found"
+  "message.no-parameters": "No parameters found",
+  "message.no-sensors-selected": "No sensors selected"
```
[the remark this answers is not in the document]
jmbrunskill: It's pretty messy right now probably should create a coldchain one at least...

frontend/packages/system/src/Notifications/api/operations.graphql
jmbrunskill: > oh didn't know that you don't need to select what you want to return does this just mean you return all fields? What about recursive queries?

The backend query just returns a string, so this seems to work. Is a bit weird though...

jmbrunskill: > Nit: just name it `runSqlQueryWithParameters`? Got confused that there actually is a query for cold chain sensors...

I think you are right, I was thinking if it had it's own name, we could more easily update it to a plugin query if needed, but I think it's not needed given we have `useColdChainSensors.ts` which is where we'd update the logic...

frontend/packages/system/src/Notifications/api/hooks/useColdChainSensors.ts
```diff
@@ -0,0 +1,37 @@
...
+    });
+
+    const sensors: SensorData[] = JSON.parse(
+      response?.runSqlQueryWithParameters
+    );
```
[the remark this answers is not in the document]
jmbrunskill: Thanks, I'll return a empty array in this case.

frontend/packages/common/src/intl/locales/en/system.json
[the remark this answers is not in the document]
jmbrunskill: https://github.com/openmsupply/notify/issues/193

frontend/packages/system/src/Notifications/api/operations.graphql
jmbrunskill: https://github.com/openmsupply/notify/pull/169/commits/8b5e154585145be482e4b80ae60846a30e20cace

frontend/packages/system/src/Notifications/api/hooks/useColdChainSensors.ts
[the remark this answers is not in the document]
jmbrunskill: https://github.com/openmsupply/notify/pull/169/commits/8b5e154585145be482e4b80ae60846a30e20cace

frontend/packages/system/src/Notifications/Pages/ColdChain/CCNotificationEditForm.tsx
jmbrunskill: https://github.com/openmsupply/notify/pull/169/commits/c9c12eb3cdd2833655fe5b7650c57f1213b72fa5

jmbrunskill:
> One thing I noticed: `send_high_temperature_alert_telegram` could be moved into the test, right?

This changes quite a lot in the next PR. will leave that for review there...

jmbrunskill:
> Do we need backend plugins at all? If you do, it would make more sense to provide the sensor list through the backend though(?) But what would be the generic interface to communicate between frontend and backend plugins? the frontend plugin sends a json blob to its backend plugin?

Yes, well we need them in that we want to support coldchain and scheduled reports and potentially other things in the future. Right now I'm just trying to implement them in their own crates until they are properly `plugins`. Maybe it is ok how it is though?

The JSON blob idea isn't crazy, hadn't thought of that. Something like `pluginDataRequest(plugin_name: String, requestBody: String)`?

I think I'll leave it for this PR though...

PR #175, opened 2023-10-05
171 cold chain should send high and low temperature alerts

### LINKED ISSUE #171: Cold chain should send high and low temperature alerts

The Send High temperature Alerts & Low Temperature Alerts checkboxes should activate notifications to configured recipients

image

High temperature alerts should look something like this...
image

## Acceptance Criteria

- [ ] High temperature alert is sent when temperature exceeds the limit configured in mSupply 
- [ ] High temperature alert doesn't repeat until the temperature has first gone down below limit
- [ ] Low temperature alert is sent when temperature is less than the limit configured in mSupply 
- [ ] Low temperature alert doesn't repeat until the temperature has first gone back above limit

## Additional context

Other features such as Ok notifications will be enabled in future issues

jmbrunskill:
Closes #171

# 👩🏻‍💻 What does this PR do?
 

Sends high or low temperature alerts something like this...

image

- [ ] High temperature alert is sent when temperature exceeds the default limit
- [ ] High temperature alert doesn't repeat until the temperature has first gone down below limit
- [ ] Low temperature alert is sent when temperature is less than the default limit
- [ ] Low temperature alert doesn't repeat until the temperature has first gone back above limit

It's a wee bit hard to test this as we need to test things like these, and any others that can be thought of...

What happens if a sensor is in the bad state before we configure it?
- Alert should be sent

What happens when the temperature goes from too low to too high?
- Alert should be sent

What happens if the same sensor is in multiple configurations?
- Each configuration is independant, so any alerts will trigger in both configuration, and the state will be stored and saved for each

What happens if the same sensor is in multiple configurations and a user is configured to get both alerts?
- Each configuration is independant, so the user will get two alerts for the same sensor (unfortunately)

Ideally we'd put in test cases in the code for these (and any other scenarios that need to be covered)

https://github.com/openmsupply/notify/issues/176 should make this possible.
Or we could refactor the code to make more of it testable without relying on the postgres database so much

You can potentially test situations with a combination of inserting records into your postgres database, e.g. something like this...

`INSERT INTO public.temperature_log(
	id, temperature, date, "time", location_id, temperature_breach_id, store_id, sensor_id, log_interval)
	VALUES ('10', 9.0000021, '2023-10-09', '15:54:00', null, null, '0E66256A14DF53478E9D872C322827E4', '454dc232-dcbb-49dc-971c-247a135eef56', 60);`

To simulate starting states, you can delete records from the `plugin_store` table in the sqlite database.


backend/coldchain/src/process.rs
```diff
@@ -0,0 +1,231 @@
...
+pub fn process_coldchain_alerts(
+    ctx: &ServiceContext,
+    current_time: NaiveDateTime,
+) -> Result<usize, ColdChainError> {
+    log::info!(
+        "Processing cold_chain configurations due at {}",
+        current_time
+    );
+
+    // Check if any cold chain configurations are due to be processed
```
jmbrunskill, line 27: Since we aren't setting a next due time, this is currently picking up all configs on every `tick`

backend/coldchain/src/process.rs
```diff
@@ -0,0 +1,231 @@
...
+                    sensor_status
+                );
+
+                let sensor_state = SensorState {
+                    sensor_id: sensor_id.clone(),
+                    status: sensor_status.clone(),
+                    timestamp: now,
+                };
+
+                let state_json = sensor_state.to_json_string()?;
```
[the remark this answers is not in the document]
andreievg: https://github.com/openmsupply/notify/compare/171-cold-chain-should-send-high-and-low-temperature-alerts...171-cold-chain-should-send-high-and-low-temperature-alerts(Example-of-get-json-value)

backend/service/src/datasource/mod.rs
```diff
@@ -17,6 +17,7 @@ pub trait DatasourceServiceTrait: Send + Sync {
         &self,
         sql_query: String,
     ) -> Result<Vec<BasicRecipientRow>, DatasourceServiceError>;
+    fn get_connection_pool(&self) -> DatasourcePool;
```
andreievg, line 20: maybe just get connection ?

andreievg (commented):
Thanks James, made a few comments, keen to chat

backend/service/src/datasource/mod.rs
jmbrunskill: I tried this but would need to import the diesel types from the datasource crate, started to get messy.
Agree it's not ideal! Maybe there's a nice way to do it?

backend/coldchain/src/process.rs
```diff
@@ -0,0 +1,231 @@
...
+            last_run_datetime: Some(now),
+            ..notification_config.clone()
+        })
+        .map_err(|e| ColdChainError::InternalError(format!("{:?}", e)))?;
+
+    let high_temp_threshold: f64 = 22.0; // TODO: Get this from config
+    let low_temp_threshold: f64 = 20.0; // TODO: Get this from config
+
+    // Loop through checking the current status for each sensor
+    for sensor_id in config.sensor_ids {
```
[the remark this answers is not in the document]
jmbrunskill: Yes, not the id itself but you configure sensos based on name, location, store

backend/coldchain/src/process.rs
```diff
@@ -0,0 +1,476 @@
...
+    NotificationConfigRowRepository::new(&ctx.connection)
+        .set_last_run_by_id(&notification_config.id, now, None)
+        .map_err(|e| ColdChainError::InternalError(format!("{:?}", e)))?;
+
+    let high_temp_threshold: f64 = config.high_temp_threshold;
+    let low_temp_threshold: f64 = config.low_temp_threshold;
+    let max_age = chrono::Duration::hours(24); // TODO: Get this from config https://github.com/openmsupply/notify/issues/179
+
+    // Put all the alerts into a vector, to simply the logic for sending alerts
+    let mut alerts: Vec<TemperatureAlert> = Vec::new();
```
jmbrunskill, line 83: I'm expecting this will change to an enum in the next PR, to handle different types of alerts....

jmbrunskill: I'm using a `vec` as we have plans to prevent too many alerts triggering all at once from a single config.
This gives us a mechanism to handle that after calculating all the alerts...

backend/coldchain/src/process.rs
```diff
@@ -0,0 +1,476 @@
...
+            .get()
+            .map_err(|e| ColdChainError::InternalError(format!("{:?}", e)))?;
+
+        let latest_temperature_row = latest_temperature(&mut connection, sensor_id.clone())
+            .map_err(|e| {
+                ColdChainError::InternalError(format!(
+                    "Failed to get latest temperature for sensor {}: {:?}",
+                    sensor_id, e
+                ))
+            })?;
```
jmbrunskill, line 102: There's quite a few things like this in the code. I'm not sure the right response.
If we can't connect to the DB, I guess skipping this iteration is probably fine?

backend/coldchain/src/sensor_state.rs
```diff
@@ -0,0 +1,124 @@
...
+pub enum SensorStatus {
+    #[default]
+    Ok,
+    LowTemp,
+    HighTemp,
+    NoData,
+}
+
+impl SensorState {
+    pub fn from_string(json_string: &str) -> Result<Self, ColdChainError> {
```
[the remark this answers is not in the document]
jmbrunskill: I think is still is, unless I do your json_serde thing in the datasource. I think I'll leave it how it is for now, and can always refactor with the other approach.

backend/coldchain/src/process.rs
```diff
@@ -0,0 +1,481 @@
...
+    let config = ColdChainPluginConfig::from_string(&notification_config.configuration_data)?;
+
+    // Update the last_checked time
+    NotificationConfigRowRepository::new(&ctx.connection)
+        .set_last_run_by_id(&notification_config.id, now, None)
+        .map_err(|e| ColdChainError::InternalError(format!("{:?}", e)))?;
+
+    let high_temp_threshold: f64 = config.high_temp_threshold;
+    let low_temp_threshold: f64 = config.low_temp_threshold;
+    let max_age = chrono::Duration::hours(24); // TODO: Get this from config https://github.com/openmsupply/notify/issues/179
```
andreievg, line 81: not_data_interval ? (as Duration)

backend/coldchain/src/process.rs
```diff
@@ -0,0 +1,481 @@
...
+
+        let sensor_status = evaluate_sensor_status(
+            now,
+            latest_temperature_row.clone(),
+            high_temp_threshold,
+            low_temp_threshold,
+            max_age,
+        );
+
+        let current_temp: String = match latest_temperature_row.clone() {
```
andreievg, line 113: Can we have temperature log rows with no temperature ?

backend/coldchain/src/process.rs
```diff
@@ -0,0 +1,481 @@
...
+            "Sensor {} is currently {:?} with a temperature of {}",
+            sensor_id,
+            sensor_status,
+            current_temp
+        );
+
+        // We need this sensor status to be unique per notification config, so we include the notification config id in the key
+        // This means that the same sensor can alarm in tow different configs
+        // And duplicate notifications would be sent, e.g. if your email address is in two configuration & you have the same sensor in both
+        // Future deduplication efforts could be considered for this...
```
andreievg, line 131: Might be good to note that the challenge is that same sensor could be in different notificaiton configs, with different settings

backend/coldchain/src/process.rs
```diff
@@ -0,0 +1,481 @@
...
+        };
+
+        log::info!(
+            "Sensor {} is currently {:?} with a temperature of {}",
+            sensor_id,
+            sensor_status,
+            current_temp
+        );
+
+        // We need this sensor status to be unique per notification config, so we include the notification config id in the key
```
andreievg, line 128: I think it's good to explain why we need this uniquenss, at this stage of reading the code I don't actually know why this is needed, maybe it becomes clear after futher reading. 
Oki, it become clear a couple of lines down, it's what status is stored against in plugin data

backend/coldchain/src/process.rs
```diff
@@ -0,0 +1,481 @@
...
+            .plugin_service
+            .get_value(ctx, PLUGIN_NAME.to_string(), sensor_status_key.clone())
+            .map_err(|e| {
+                ColdChainError::InternalError(format!(
+                    "Failed to get previous state for sensor {}: {:?}",
+                    sensor_id, e
+                ))
+            })?;
+
+        let prev_sensor_status = match prev_sensor_status {
```
andreievg, line 146: I still think that my suggestions is worth doing, where this logic is done by the plugin_service. There is an example https://github.com/openmsupply/notify/pull/175#discussion_r1347317445

backend/coldchain/src/process.rs
```diff
@@ -0,0 +1,481 @@
...
+
+        log::info!(
+            "Status for sensor {} has changed from {:?} to {:?}",
+            sensor_id,
+            prev_sensor_status,
+            sensor_status
+        );
+
+        // Persist the new status
+        // Note: Since we only persist `new` statuses, if a sensor has always been in the ok state, it won't have a record in the plugin store
```
andreievg, line 200: nice comment

backend/coldchain/src/sensor_info.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub store_id: String,
+    #[diesel(sql_type = Text)]
+    pub location_name: String,
+    #[diesel(sql_type = Text)]
+    pub sensor_name: String,
+    #[diesel(sql_type = Nullable<Double>)]
+    pub batterylevel: Option<f64>,
+}
+
+pub fn sensor_info(
```
andreievg, line 25: Do you think it's worth using diesel here ? vs raw sql ?

backend/coldchain/src/process.rs
```diff
@@ -0,0 +1,481 @@
...
+
+    // look up the recipients for the notification config
+    let notification_targets =
+        get_notification_targets(ctx, &notification_config).map_err(|e| {
+            ColdChainError::InternalError(format!("Failed to get notification targets: {:?}", e))
+        })?;
+
+    for alert in alerts {
+        // Send the notifications
+        let result = send_temperature_alert(
```
andreievg, line 312: maybe should be queue_temperature_alert

backend/coldchain/src/process.rs
```diff
@@ -0,0 +1,481 @@
...
+
+        let sensor_row = match sensor_row {
+            Some(row) => row,
+            None => {
+                log::error!("No sensor info found for sensor {}", sensor_id);
+                continue;
+            }
+        };
+
+        let alert = match sensor_status {
```
andreievg, line 248: I like this match statement

backend/coldchain/README.md
```diff
@@ -0,0 +1,28 @@
...
+## Installation
+
+Configure notify to use a postgres database that's receiving cold chain data from mSupply.
+
+## Logic
+
+Here's the state diagram summarising the logic of the cold chain alerts.
+
+```mermaid
+flowchart TD
```
andreievg, line 15: I think this diagram should have a bit more (I think when full logic is implemented), like getting and saving state, calculating new state

andreievg (approved):

andreievg (approved):
Looks great.

Personally when processing multiple records like this I prefer this pattern:

**main method**


for each notification config
  for each sensor
    match  try process sensor notification
       error => log
        ok => add it to array

process array (queue it, dedup it, truncate it)


**try process sensor notification**

We don't need to log it here, and steps become less verbose (consdiering this will be a big method, and it already is, I think it a good idea to make it slightly less verbose)

i.e. 

```diff
diff --git a/backend/coldchain/src/process.rs b/backend/coldchain/src/process.rs
index 0cc6abc..b98f989 100644
--- a/backend/coldchain/src/process.rs
+++ b/backend/coldchain/src/process.rs

@@ -237,13 +237,10 @@ fn try_process_coldchain_notifications(
             ))
         })?;
 
-        let sensor_row = match sensor_row {
-            Some(row) => row,
-            None => {
-                log::error!("No sensor info found for sensor {}", sensor_id);
-                continue;
-            }
-        };
+        let sensor_row = sensor_row.ok_or(ColdChainError::Other(format!(
+            "No sensor info found for sensor {}",
+            sensor_id
+        )))?;
 
         let alert = match sensor_status {
             SensorStatus::HighTemp => match config.high_temp {

```

or even better, if there is an error variant that implements form for anyhow:

```diff
diff --git a/backend/coldchain/src/process.rs b/backend/coldchain/src/process.rs
index 0cc6abc..a7dc3b9 100644
--- a/backend/coldchain/src/process.rs
+++ b/backend/coldchain/src/process.rs
 
@@ -237,13 +237,8 @@ fn try_process_coldchain_notifications(
             ))
         })?;
 
-        let sensor_row = match sensor_row {
-            Some(row) => row,
-            None => {
-                log::error!("No sensor info found for sensor {}", sensor_id);
-                continue;
-            }
-        };
+        let sensor_row =
+            sensor_row.with_context(|| format!("No sensor info found for sensor {}", sensor_id))?;
 
         let alert = match sensor_status {
             SensorStatus::HighTemp => match config.high_temp {

```

andreievg (approved):
Looks great.

Personally when processing multiple records like this I prefer this pattern:

**main method**


for each notification config
  for each sensor
    match  try process sensor notification
       error => log
        ok => add it to array

process array (queue it, dedup it, truncate it)


**try process sensor notification**

We don't need to log it here, and steps become less verbose (consdiering this will be a big method, and it already is, I think it a good idea to make it slightly less verbose)

i.e. 

```diff
diff --git a/backend/coldchain/src/process.rs b/backend/coldchain/src/process.rs
index 0cc6abc..b98f989 100644
--- a/backend/coldchain/src/process.rs
+++ b/backend/coldchain/src/process.rs

@@ -237,13 +237,10 @@ fn try_process_coldchain_notifications(
             ))
         })?;
 
-        let sensor_row = match sensor_row {
-            Some(row) => row,
-            None => {
-                log::error!("No sensor info found for sensor {}", sensor_id);
-                continue;
-            }
-        };
+        let sensor_row = sensor_row.ok_or(ColdChainError::Other(format!(
+            "No sensor info found for sensor {}",
+            sensor_id
+        )))?;
 
         let alert = match sensor_status {
             SensorStatus::HighTemp => match config.high_temp {

```

or even better, if there is an error variant that implements form for anyhow:

```diff
diff --git a/backend/coldchain/src/process.rs b/backend/coldchain/src/process.rs
index 0cc6abc..a7dc3b9 100644
--- a/backend/coldchain/src/process.rs
+++ b/backend/coldchain/src/process.rs
 
@@ -237,13 +237,8 @@ fn try_process_coldchain_notifications(
             ))
         })?;
 
-        let sensor_row = match sensor_row {
-            Some(row) => row,
-            None => {
-                log::error!("No sensor info found for sensor {}", sensor_id);
-                continue;
-            }
-        };
+        let sensor_row =
+            sensor_row.with_context(|| format!("No sensor info found for sensor {}", sensor_id))?;
 
         let alert = match sensor_status {
             SensorStatus::HighTemp => match config.high_temp {

```

backend/coldchain/src/process.rs
andreievg: I think it's fine to skip, I have an overall suggestion in the PR comment, that would only be ok if we skip

backend/coldchain/src/process.rs
andreievg: Sounds good

backend/coldchain/src/alerts.rs
```diff
@@ -29,29 +29,27 @@ Temperature: 10° C
...
-    pub location_id: String,
-    pub location_name: String,
     pub sensor_id: String,
     pub sensor_name: String,
+    pub store_name: String,
+    pub location_name: String,
     pub datetime: NaiveDateTime,
-    pub temperature: f64,
+    pub temperature: String,
+    pub alert_type: String,
```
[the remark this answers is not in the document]
jmbrunskill: Done here (different PR): https://github.com/openmsupply/notify/pull/189/files#diff-6d4438989415197ad6077df1f06ddcf3cbbb50a93247f496a84f97008e0b185b

backend/coldchain/src/latest_temperature.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+pub fn latest_temperature(
+    connection: &mut PgConnection,
+    sensor_id: String,
+) -> Result<Option<LatestTemperatureRow>, DieselError> {
+    let query = "SELECT 
+    id,
+    sensor_id,
+    CONCAT(TO_CHAR(date,'YYYY-MM-DD'),' ', TO_CHAR(time,'HH24:MI:SS'))::timestamp AS log_datetime,
```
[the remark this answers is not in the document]
jmbrunskill: The where clause is running on the date and time fields so I don't think this is a performance problem, this is how the conversion is done for dashboard queries so thought it was good to be consistent.

The fact that date and time are not indexed on the other hand, that *IS* a problem!
Have a discussion going with Chris and Adam about this, and other potential performance problems with large installations, will keep you in the loop.

backend/coldchain/src/parse.rs
```diff
@@ -0,0 +1,143 @@
...
+    #[serde(default = "default_low_temp_limit")]
+    pub low_temp_threshold: f64,
+    #[serde(default = "default_high_temp_limit")]
+    pub high_temp_threshold: f64,
+    #[serde(default)]
+    pub no_data: bool,
+    #[serde(default = "default_no_data_interval")]
+    pub no_data_interval: u32,
+    #[serde(default = "default_no_data_units")]
+    pub no_data_units: String,
```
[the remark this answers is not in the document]
jmbrunskill: Good catch, it's in the frontend too annoyingly...

backend/coldchain/src/sensor_info.rs
jmbrunskill: I guess we could? Doesn't seem worth changing right now...

backend/coldchain/src/process.rs
jmbrunskill: Possibly?

image


It's annoying to add the checks in. Maybe I should assume it's always populated. Was trying to be defensive as I don't know what we might find out there :)

backend/coldchain/src/process.rs
jmbrunskill: Thanks! Feels like a lot of duplication and I've just noticed a bug with   `datetime: Local::now().naive_local()` only in one match though...
I agree the concept is good, helps break up the problem of what alert to create and evaluate if that type of alert should even be created...

jmbrunskill:
> Personally when processing multiple records like this I prefer this pattern:
> 
> **main method**
> 
> for each notification config for each sensor match try process sensor notification error => log ok => add it to array
> 

I agree, have created a new issue for this. Mainly because I have too many branches now and trying to refactor here would probabLy create too much chaos!

backend/coldchain/src/process.rs
jmbrunskill: Yes it should be! See - https://github.com/openmsupply/notify/pull/189/files#diff-29abfdb7ead8b1acb12b6e2963ad9c4e40af3012cb8ef74d5e1d6cedf5235f5eR81

PR #182, opened 2023-10-09
Add a build-windows command to yarn for jenkins to use

### LINKED ISSUE #178: Fix version number on Jenkins Build
## Describe the bug

image

## To Reproduce

1. Build with Jenkins
2. Install
3. Check Admin Page for Version

## Expected behaviour

Tagged Build version should show up...

jmbrunskill:
Closes #178

# 👩🏻‍💻 What does this PR do?
 

Passes the version number from jenkins build for windows into the webpack process so it can be shown in the UI.

Tested in Jenkins.

I tried various things, and tired to read about how to use [cross-env](https://www.npmjs.com/package/cross-env) but after wasting too much time went with the different build target approach.
Happy for someone else to make it more consistent!

### DIFF
--- a/build_windows.bat
+++ b/build_windows.bat
@@ -1,0 +1,14 @@
+IF NOT EXIST version.txt (
+    ECHO ##### No version.txt found, defaulting to package version #####
+    node -p "require('./frontend/package.json').version" > version.txt
+)
+
+ECHO ##### Prepare notify frontend #####
+set /p APP_BUILD_VERSION=<version.txt
+echo %APP_BUILD_VERSION%
+cd "frontend" && yarn install --force --frozen-lockfile && yarn build-windows
+if %errorlevel% neq 0 exit /b %errorlevel%
+
+ECHO ##### Building notify backend #####
+cd ../backend && cargo build --release && cargo build --release --bin notify_service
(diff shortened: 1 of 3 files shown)


mark-prins (approved):
We've used the version from the root `package.json` rather than using an env var. It doesn't automatically update from the git tag though, and that's a nice touch.

Disappointingly, `cross-env` failed to deliver - it did update the env var to the text `%APP_BUILD_VERSION%` on windows, but didn't pick up the var.

I would use a batch file rather than having all the commands in jenkins, as it would make it easier to see where the env var is being set - but what we have here will work fine

jmbrunskill:
> Disappointingly, `cross-env` failed to deliver - it did update the env var to the text `%APP_BUILD_VERSION%` on windows, but didn't pick up the var.

Ah gutted, I couldn't get it to work either. Not sure what I'm doing wrong. Something to do with the layers of yarn, webpack, lerna, etc perhaps? I started to look at a few things such dotenv as well which might work, but like you agree it's not worth wasting time on right now!

> I would use a batch file rather than having all the commands in jenkins, as it would make it easier to see where the env var is being set - but what we have here will work fine

Good point I effectively have that for linux build, will do the same for the windows one...

PR #184, opened 2023-10-09
183 cold chain alert date time should be shown in server time

### LINKED ISSUE #183: Cold Chain Alert Date Time - Should be shown in server time
## Describe the bug

image

## To Reproduce

1. Send a cold chain alert, compare time specified to local time on server

## Expected behaviour

Alert should be based on server local time (user timezone isn't available)

## Additional context

The time is stored in UTC internally so will need to be converted to another timezone...

jmbrunskill:
Closes #183

# 👩🏻‍💻 What does this PR do?
 

What it says on the box! Server local time is passed to the alert.
image

Only tested locally with my timezone so far.

To test, create a Cold Chain Alert and se what time is displayed

Do we need to do more here? Maybe display the actual timezone?

We could allow the user to specify a timezone for alerts?
Pass both UTC and Local time to the template incase people want to do something custom?

### DIFF
--- a/backend/coldchain/src/process.rs
+++ b/backend/coldchain/src/process.rs
@@ -1,4 +1,4 @@
-use chrono::NaiveDateTime;
+use chrono::{Local, NaiveDateTime};
 use repository::{
     NotificationConfigKind, NotificationConfigRowRepository, NotificationConfigStatus,
 };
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
Change seems reasonable to me. I'm having trouble setting up, so have not tested though.

PR #188, opened 2023-10-10
185 configuable temperature limits for cold chain

### LINKED ISSUE #185: Configuable temperature limits for Cold Chain

Usually cold chain notification should use standardise limits for temperature limits between 2-8 degrees C

However for testing and other possible use cases, configurable limits would be useful

The default 2 - 8 limit should be on by default, but give the user an option to update/change these limits if they will

## Acceptance Criteria

- [ ] defaults are visible and easy to configure
- [ ] custom limits can be specified (for all sensors in the config)
- [ ] Update `i` icon messages to explain the default limits

- [ ] Update mSupply & coldchain apps to store & sync an excursion limit (rather than breach limits)
- [ ] Some kind of default limits specified in the server config?

## Additional context

Breach limits may not be synced to mSupply (apparently)
Breach limits are setup to have a duration (not a single excusion alarm)

jmbrunskill:
Closes #185

# 👩🏻‍💻 What does this PR do?
 

image

Acceptance criteria
- [ ]  defaults are visible and easy to configure
- [ ]  custom limits can be specified (for all sensors in the config)
- [ ]  Update i icon messages to explain the default limits

- Update the limits and check that they are saved correctly.
- Check that limits are disabled when not enabled
- Check defaults are correctly set

No backend changes were needed for this, as it was already handled correctly in the backend.


mark-prins:
It might be helpful to have the two limits affect each other to prevent misconfiguring like so:

image

frontend/packages/common/src/intl/locales/en/system.json
```diff
@@ -70,7 +70,7 @@
   "messages.recipients-added_other": "{{count}} recipients added to list",
   "messages.recipients-added_one": "{{count}} recipient added to list",
   "messages.cold-chain-no-data-information": "Alert will be triggered if no temperature data is received by the server within the specified timeframe",
-  "messages.cold-chain-temperature-information": "Temperature thresholds are configured in the mSupply Cold Chain application",
+  "messages.cold-chain-temperature-information": "Default Temperature thresholds are 2-8 degrees Celsius for cold chain",
```
mark-prins, line 73: It's not clear what 'for cold chain' means here, even in context. Do you mean the cold chain application? 
Regardless, I'm wondering if it matters. It's enough to say that the default is 2 & 8°C
```suggestion

-   "messages.cold-chain-temperature-information": "Default Temperature thresholds are 2°C and 8°C",

```

frontend/packages/system/src/Notifications/Pages/ColdChain/CCNotificationEditForm.tsx
```diff
@@ -50,15 +51,27 @@ export const CCNotificationEditForm = ({
...
+
           <label htmlFor="highTemp">
             {t('label.coldchain-high-temp-alerts')}
-            <Tooltip title={t('messages.cold-chain-temperature-information')}>
-              <span>
-                {' '}
-                <InfoIcon fontSize="small" color="inherit" />
-              </span>
-            </Tooltip>
+            {'  '}
```
mark-prins, line 57: It's unusual to use spacing like this, but I've left it in place. Feels like more work to replace with layout components 🤷

frontend/packages/system/src/Notifications/Pages/ColdChain/CCNotificationEditForm.tsx
```diff
@@ -50,15 +51,27 @@ export const CCNotificationEditForm = ({
...
           </label>
+          <NumberInput
+            id={'highTempThreshold'}
+            disabled={!draft.highTemp}
+            value={draft.highTempThreshold}
+            required
+            onChange={newValue => onUpdate({ highTempThreshold: newValue })}
+            sx={{ width: '60px' }}
+          />
+          <Tooltip title={t('messages.cold-chain-temperature-information')}>
```
mark-prins, line 67: The info icons look too big to me.

I've changed to this, by replacing with the InfoTooltipIcon:

image

This requires adding `display: 'inline-flex'` to the div which is used by that component:

```
export const InfoTooltipIcon = ({ title }: { title: string }) => (
  <Tooltip title={title}>
    <div
      style={{
        transform: 'scale(0.7)',
        cursor: 'help',
        display: 'inline-flex',
      }}
    >
      <InfoIcon fontSize="small" />
    </div>
  </Tooltip>
```

changes in this [suggestion branch](https://github.com/openmsupply/notify/compare/185-configuable-temperature-limits-for-cold-chain...185-info-icon-suggestion) 😉 

Added bonus is that the component has the cursor showing as a ? which is neat too

mark-prins (approved):
👍  Lovely - that's working nicely. As usual, I've popped in a UI suggestion..

:nit The '1 Minutes' and '1 Hours' could be more elegant. Even just changing to 'Minute(s)' 🤔 

image

frontend/packages/system/src/Notifications/Pages/ColdChain/CCNotificationEditForm.tsx
jmbrunskill: Nice, InfoTooltipIcon could be a good common component perhaps

jmbrunskill:
> 👍 Lovely - that's working nicely. As usual, I've popped in a UI suggestion..
> 
> :nit The '1 Minutes' and '1 Hours' could be more elegant. Even just changing to 'Minute(s)' 🤔
> 
> image

Kapture 2023-10-25 at 16 02 44

Probably overkill but why not :)

frontend/packages/system/src/Notifications/Pages/ColdChain/CCNotificationEditForm.tsx
jmbrunskill: I agree it's not normal, seemed a like a good enough quick fix for now.

jmbrunskill:
> It might be helpful to have the two limits affect each other to prevent misconfiguring like so:
> 
> image

image

Added an error in this situation. No explanation text but again, I think as long as we call attention to it, we should be ok.

PR #187, opened 2023-10-10
186 cold chain alerts should default to enabled

jmbrunskill:
Closes #186

# 👩🏻‍💻 What does this PR do?
 

Default to all alerts turned on for Cold Chain

image

# 🧪 How has/should this change been tested?

1. Create a new cold chain alert.
1. Check the defaults are correct
1. Check that the existing configurations haven't be inadvertently updated.

## 💌 Any notes for the reviewer?

I've also removed the backwards compatibility code for recipientIds, as I think the only placing which was still using this old method has been updated now.

PR #189, opened 2023-10-10
179 cold chain no data alert

### LINKED ISSUE #179: Cold Chain - No Data Alert

When no data is available for a sensor and the Send No Data Alert is enabled in cold chain, send a notification message.

image

## Acceptance Criteria

Assuming **Send 'no temperature data' alert after** is set to 1 hour

- [ ] If the last state was not `NoData` and there's a null temperature in the latest database row (Send Alert)
- [ ] If the last state was not `NoData` and there's no data returned from the sensor (aka no database row) (Send Alert)
- [ ] If the last state was not `NoData` and the time stamp from the latest database row is more than 1 hour old (Send Alert)

jmbrunskill:
Closes #179

# 👩🏻‍💻 What does this PR do?
 

Sensor that's never had data:
image

Sensor that's has data but it's old...

image

Should the date also show the time from the last record rather than alert time? I think probably yes?

For bonus points, I've also implemented the `Recovered` or `Confirm Ok` messages as well.

We need to translate minutes into hours and days I think.


backend/coldchain/src/parse.rs
```diff
@@ -69,7 +71,7 @@ fn default_high_temp_limit() -> f64 {
 }
 
 fn default_no_data_interval() -> u32 {
-    1
+    2
```
jmbrunskill, line 74: The default value has changed in this PR: https://github.com/openmsupply/notify/pull/187

backend/coldchain/src/alerts.rs
```diff
@@ -28,28 +28,53 @@ Temperature: 10° C
...
-        title_template: Some(TemplateDefinition::TemplateName(
+    let title_template = match alert.alert_type {
+        AlertType::High => Some(TemplateDefinition::TemplateName(
             "coldchain/temperature_title.html".to_string(),
         )),
-        body_template: TemplateDefinition::TemplateName("coldchain/temperature.html".to_string()),
+        AlertType::Low => Some(TemplateDefinition::TemplateName(
+            "coldchain/temperature_title.html".to_string(),
+        )),
+        _ => None,
```
andreievg, line 64: Do we need a title for no data alert ?

backend/coldchain/src/process.rs
```diff
@@ -118,6 +118,14 @@ fn try_process_coldchain_notifications(
             None => "Never Recorded".to_string(),
         };
 
+        let data_age: String = match latest_temperature_row.clone() {
+            Some(row) => format!(
+                "{} minutes",
+                (now - row.log_datetime).num_minutes().to_string() // TODO: Improve this to show the age in hours/days/weeks/months/years? Ideally it would be translatable?
```
andreievg, line 124: Yeah would be great to also have it translatable !

andreievg:
> Should the date also show the time from the last record rather than alert time? I think probably yes?

I think you should show both really, but clearly differentiated (notification date/time and last record received date/time)

backend/coldchain/src/process.rs
```diff
@@ -247,14 +255,15 @@ fn try_process_coldchain_notifications(
 
         let alert = match sensor_status {
             SensorStatus::HighTemp => match config.high_temp {
-                true => Some(TemperatureAlert {
+                true => Some(ColdchainAlert {
                     store_name: sensor_row.store_name.clone(),
```
andreievg, line 259: I like the idea of using base cold chain alert here, I think it makes it easier to see applied alert difference and easier for future modifications

andreievg (approved):

andreievg:
Looks good thanks ! 

I think considering the branching logic, you've made it quite readable !

Don't want to complicate things in logic, but had a question about high temperature alert when no data is not checked but high/low alert notification is checked, I don't want to complicate things so maybe it's best to clearly explain to user when a particular alert will be triggered. Otherwise you need a collection of alerts for a sensor and then more complex logic when deducing the alert, or a stackable alert (with both notifications as one)

backend/coldchain/src/alerts.rs
jmbrunskill: Yeah probably should

jmbrunskill:
> Don't want to complicate things in logic, but had a question about high temperature alert when no data is not checked but high/low alert notification is checked, I don't want to complicate things so maybe it's best to clearly explain to user when a particular alert will be triggered. Otherwise you need a collection of alerts for a sensor and then more complex logic when deducing the alert, or a stackable alert (with both notifications as one)

I like the simplicity of this approach but agree there are a few rough edges. Not sure if it's what you're describing but if (for example) we have a high temperature, then we get no data then we get high temperature again, and the user has turned off no-data alerts, we'd send 2 high temperture alerts I think.

PR #191, opened 2023-10-10
Remove manual PartialEq implementations

### LINKED ISSUE #190: Remove unneccesary PartialEq Implementations
## Describe the bug

We have a lot of code like this...
```
// PartialEq is only needed for tests
impl PartialEq for ModifyNotificationConfigError {
    fn eq(&self, other: &Self) -> bool {
        match (self, other) {
            (
                ModifyNotificationConfigError::NotificationConfigAlreadyExists,
                ModifyNotificationConfigError::NotificationConfigAlreadyExists,
            ) => true,

            (
                ModifyNotificationConfigError::ModifiedRecordNotFound,
                ModifyNotificationConfigError::ModifiedRecordNotFound,
            ) => true,
```

Often this can be replaced with `#[derive(PartialEq)] which is safer and less code

## Additional context

The manually implemented code is a potential risk to the code base.

jmbrunskill:
Closes #190 

# 👩🏻‍💻 What does this PR do?
 

Removes some PartialEq implementations, that weren't needed and were causing bad copy and pasting behaviour :)

No testing is required, rust compiles!

### DIFF
--- a/backend/graphql/user_account/src/mutations/create.rs
+++ b/backend/graphql/user_account/src/mutations/create.rs
@@ -94,8 +94,7 @@ fn map_error(error: ModifyUserAccountError) -> Result<CreateUserAccountResponse>
         }
         ModifyUserAccountError::DatabaseError(_) => InternalError(formatted_error),
         ModifyUserAccountError::ModifiedRecordNotFound => InternalError(formatted_error),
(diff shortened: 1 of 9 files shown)


clemens-msupply (approved):
Looks good!

Issue #65, opened 2023-08-21
Test method for Scheduled Reports


jmbrunskill:
## Is your feature request related to a problem? Please describe

Right now it's really hard to test out a scheduled report to see if it works, it would be great if there's a way to 

## Describe the solution you'd like

Add a new `Test` button similar to the test query button on the Notification Query Page.

It should return a view with information such as.

- [ ] How the template would render, or the error received trying to render the template
- [ ] List of recipients that would be notified
- [ ] Any other errors that can occur would be available

## Describe alternatives you've considered

- [x] A run this now button would be very useful
- [x] A veiw allowing you to see all notifications, espeically any failed ones with their errors would be useful

PR #195, opened 2023-10-12
137 schedule notifications should load data from sql queries to render in the template

### LINKED ISSUE #137: Schedules Notifications should Load data from sql queries to render in the template

Currently when sending a scheduled notification, if any sql queries are configured, they aren't run which means they aren't available to the rendered template. 

When rendering a scheduled notification the data retrieved from each query configured should be available to the template by it's `reference name`

jmbrunskill:
Closes #137 

# 👩🏻‍💻 What does this PR do?
 

Runs any configured queries and makes their data available in the rendering of the template.

Create a query which returns known data.
Create a notification config for a scheduled report

Add you query and update the body template to include data based on the reference name.

Could be something like this...

Query (last_10_temperature_logs):
```
SELECT s.name, tl.temperature, "time", "date"
FROM temperature_log tl
JOIN sensor s ON tl.sensor_id = s.id
ORDER BY date desc, time desc
LIMIT 10
```

Template (first record)

```
Sensor:  {{ last_10_temperature_logs.0.name }} Temperature {{ last_10_temperature_logs.0.temperature }}
```

Template all 10 records

```
{% for row in last_10_temperature_logs %}
{{row.name}}: {{row.temperature}}
{% endfor %}
```

Right now this is a nightmare to get setup and test with anything complicated, see https://github.com/openmsupply/notify/issues/65 for some ideas there.

### DIFF
--- a/backend/scheduled/src/query.rs
+++ b/backend/scheduled/src/query.rs
@@ -1,0 +1,295 @@
+use std::collections::HashMap;
+
+use repository::{EqualFilter, NotificationQueryFilter, NotificationQueryRepository};
+use serde_json::json;
+use service::{notification_config::query::NotificationConfig, service_provider::ServiceContext};
+
+use crate::{parse::ScheduledNotificationPluginConfig, NotificationError};
+
+pub fn get_notification_query_results(
+    ctx: &ServiceContext,
+    notification_config: &NotificationConfig,
+    config: &ScheduledNotificationPluginConfig,
+) -> Result<HashMap<String, serde_json::Value>, NotificationError> {
+    let mut query_results = HashMap::new();
+
+    // get all the configured queries
+
+    let notification_config_repo = NotificationQueryRepository::new(&ctx.connection);
+
+    let queries = notification_config_repo
+        .query_by_filter(NotificationQueryFilter::new().id(EqualFilter::equal_any(
+            config.notification_query_ids.clone(),
+        )))
+        .map_err(|e| {
+            NotificationError::InternalError(format!("Unable to get notification queries: {:?}", e))
+        })?;
+
+    // loop through all the notification query ids, run them, and store the results
+    for query in queries {
+        let result = ctx
+            .service_provider
+            .datasource_service
+            .run_sql_query_with_parameters(
+                query.query.clone(),
+                notification_config.parameters.clone(),
+            );
+        let query_json = match result {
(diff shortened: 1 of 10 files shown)


backend/scheduled/src/query.rs
```diff
@@ -0,0 +1,295 @@
...
+            Ok(result) => serde_json::from_str(&result)
+                .unwrap_or_else(|_| json!([{"error": "Unable to parse query result"}])),
+            Err(e) => {
+                log::error!(
+                    "Error running query {} for notification {}: {:?}",
+                    query.reference_name,
+                    notification_config.title,
+                    e
+                );
+                json!([{"error": "error running query", "query": query.query, "parameters": notification_config.parameters}])
```
jmbrunskill, line 47: Hoping this or similar will be useful for debugging. Exactly how it will be used is yet to be decided...

frontend/packages/system/src/Notifications/Pages/Base/BaseNotificationEditPage.tsx
```diff
@@ -90,11 +90,15 @@ export const BaseNotificationEditPage = <T extends BaseNotificationConfig>({
...
 
-  const requiredParams = (sqlRecipientLists?.nodes ?? [])
+  const requiredRecipientParams = (sqlRecipientLists?.nodes ?? [])
     .filter(list => draft.sqlRecipientListIds.includes(list.id))
     .map(list => list.parameters)
     .flat(1);
 
+  const requiredConfigParams = draft.requiredParameters;
+
+  const requiredParams = requiredRecipientParams.concat(requiredConfigParams);
```
jmbrunskill, line 100: This is deduped later in the Params Panel

backend/service/src/notification/enqueue.rs
```diff
@@ -129,7 +129,7 @@ pub fn create_notification_events(
             Err(e) => {
                 log::error!("Failed to render notification title template: {:?}", e);
                 NotificationEventRow {
-                    status: NotificationEventStatus::Errored,
+                    status: NotificationEventStatus::Failed,
```
jmbrunskill, line 132: This is a bug I discovered. `Errored` would still be sent (it means it can be retried) possibly should change this name to avoid future confusion...

mark-prins (approved):
👍  Working nicely!

PR #219, opened 2023-10-31
200 change template format to markdown

### LINKED ISSUE #200: Change Template Format to Markdown

Right now we have a single template for each notification, it is assumed to be in HTML format which works well for emails but doesn't work well for Telegram. To make the template acceptable to Telegram makes it useless for emails...

What should we do?

1. Use Markdown and convert to HTML for emails?
2. Use HTML then convert to markdown for telegram?
3. Use HTML then stripout invalid stuff for html going to telegram?
4. Have different templates for each notification type (gets out of hand very quickly!)

## Acceptance Criteria

- [ ] Something

jmbrunskill:
Closes #200

# 👩🏻‍💻 What does this PR do?
 

Telegram
image

HTML - email
image

Plaintext - email
image

This change is not the magic bullet I was hoping.
I still need to escape characters to avoid issues with basic characters people want to use such as `-`.
Telegram also doesn't seem to handle things like ordered lists

Perhaps I need to handle the rendering better for telegram by parsing the markdown to html, then removing unacceptable html such as `<p>`?
Or maybe we need to revert to the idea of specific templates for each channel (e.g. telegram can be different to email etc)

https://crates.io/crates/pulldown-cmark-to-cmark looks like it would allow filtering markdown and back to markdown, but would need a bit of work to figure it out, so not sure it's worth it!


mark-prins (approved):
👍 changes look good to me
still having troubles with email, so I haven't verified that's working. Telegram is sending me messages and not errors, so that's good!

Issue #153, opened 2023-09-26
Queries: Show the actual query when we click on Test SQL Query, also keep showing the error message somewhere


Mai21:
## Is your feature request related to a problem? Please describe

When we run 'Test SQL Query',  the error shows but sometimes hard to know what causes the error.
So showing the actual query will help us to know why the error happens.
Also, the error disappears after a while now, but keeps showing somewhere will be helpful too.

![Screenshot 2023-09-26 at 11 05 58 AM](https://github.com/openmsupply/notify/assets/36495540/421318e7-667d-4f7a-be99-df46f3754311)

PR #221, opened 2023-10-31
#153 queries show the actual query when we click on test sql query, also keep showing the error message somewhere

Mai21:
Closes 153

# 👩🏻‍💻 What does this PR do?
 
- Added "Generate query" in the side panel 
- Added a permanent alert panel instead of a snack bar 
- Added an information panel for the number of rows

When the query returns an error
Screenshot 2023-11-01 at 11 08 48 AM

When the query returns the result
Screenshot 2023-11-01 at 11 09 05 AM

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2


frontend/packages/common/src/ui/components/panels/InfoPanel.tsx
```diff
@@ -10,8 +10,6 @@ export const InfoPanel = ({ message }: { message: string }) => (
     alignItems="center"
     sx={{
       backgroundColor: 'gray.pale',
-      borderRadius: '10px',
```
jmbrunskill, line 13: Normally I'd be careful about changing a common components settings like this.
I have checked it's not used anywhere else though, so in this case it's fine.

frontend/packages/system/src/Queries/NotificationQueries/DetailEdit.tsx
```diff
@@ -101,9 +118,18 @@ export const DetailEdit = () => {
       </AppBarContentPortal>
       {/* Sql Results table */}
       <Box sx={{ width: '100%', display: 'flex', flexDirection: 'column' }}>
-        <Box sx={{ flex: '1', overflow: 'auto' }}>
+        {queryError && (<AlertPanel message ={queryError}/>)}
+        {(generatedQuery && !queryError) && (<InfoPanel message ={'Result: ' + sqlResults.length.toString() + ' rows'}/>)}
```
jmbrunskill, line 122: Maybe should create a translation for this? Not a priority though...

frontend/packages/system/src/Queries/NotificationQueries/QueryEditor.tsx
```diff
@@ -257,7 +257,6 @@ export const QueryEditor = ({
             onClick={() => {
               onSave(draft).catch(err => {
                 console.error(err);
-                error(err)();
```
jmbrunskill, line 260: Why remove this?

jmbrunskill: There could be an error saving which would be good to show to the user if that happens...

jmbrunskill (approved):
Looking good. Have a few suggestions for improvements to the code...

frontend/packages/common/src/ui/components/panels/InfoPanel.tsx
Mai21: I understand it:) It is not used, so I changed it. But if it shouldn't be changed, I will create a new component for this!

backend/graphql/datasource/src/lib.rs
```diff
@@ -33,8 +36,8 @@ impl DatasourceQueries {
         // TODO some kind of query validation?
 
         // Query datasource service and return result
-        match datasource_service.run_sql_query(sql_query) {
-            Ok(result) => Ok(result),
+        match datasource_service.run_sql_query_with_parameters(sql_query, "".to_string()) {
```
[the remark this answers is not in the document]
Mai21: @jmbrunskill It seems not being used but I think I changed it accidentally... I'll return this to the previous one

```Rust
match datasource_service.run_sql_query(sql_query) {
```

frontend/packages/system/src/Queries/NotificationQueries/QueryEditor.tsx
Mai21: The reason why I remove this is I removed `error` here. This is for using snackbar. 
https://github.com/openmsupply/notify/blob/be0151fb95dee12c4a1042e913e668f449453577/frontend/packages/system/src/Queries/NotificationQueries/QueryEditor.tsx#L61 

But for saving in the QueryEditor, I agree with you, we better have this error catch. I'll get it back.

frontend/packages/system/src/Queries/NotificationQueries/DetailEdit.tsx
Mai21: Done

jmbrunskill (approved):
Looks good thanks for the updates!

frontend/packages/common/src/intl/locales/en/system.json
```diff
@@ -73,6 +73,7 @@
   "messages.password-weak": "Weak",
   "messages.recipients-added_other": "{{count}} recipients added to list",
   "messages.recipients-added_one": "{{count}} recipient added to list",
+  "messages.query-result-count": "Result: {{number}} rows",
```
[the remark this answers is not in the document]
Mai21: @jmbrunskill I was wondering how to use `{{count}}` then decided to go with `{{number}}`. 
I fixed it but when the result is 0, it also returns with `0 rows`. Is there any way that we can return 0 row, 1 row or 2 rows?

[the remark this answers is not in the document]
jmbrunskill: I think how it's working is correct `0 rows` is the right way to say it.

https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/z/zero-zeros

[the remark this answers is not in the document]
jmbrunskill: Here's the library docs
https://www.i18next.com/translation-function/plurals#singular-plural
https://www.i18next.com/translation-function/plurals#languages-with-multiple-plurals

PR #227, opened 2023-11-06
213 follow up reminders for cold chain alerts

### LINKED ISSUE #199: Refactor cold chain alerts to break up logic more
Personally when processing multiple records like this I prefer this pattern:

**main method**

for each notification config
  for each sensor
    match  try process sensor notification
       error => log
        ok => add it to array

process array (queue it, dedup it, truncate it)

**try process sensor notification**

We don't need to log it here, and steps become less verbose (consdiering this will be a big method, and it already is, I think it a good idea to make it slightly less verbose)

i.e. 

```diff
diff --git a/backend/coldchain/src/process.rs b/backend/coldchain/src/process.rs
index 0cc6abc..b98f989 100644
--- a/backend/coldchain/src/process.rs
+++ b/backend/coldchain/src/process.rs

@@ -237,13 +237,10 @@ fn try_process_coldchain_notifications(
             ))
         })?;
 
-        let sensor_row = match sensor_row {
-            Some(row) => row,
-            None => {
-                log::error!("No sensor info found for sensor {}", sensor_id);
-                continue;
-            }
-        };
+        let sensor_row = sensor_row.ok_or(ColdChainError::Other(format!(
+            "No sensor info found for sensor {}",
+            sensor_id
+        )))?;
 
         let alert = match sensor_status {
             SensorStatus::HighTemp => match config.high_temp {

```

or even better, if there is an error variant that implements form for anyhow:

```diff
diff --git a/backend/coldchain/src/process.rs b/backend/coldchain/src/process.rs
index 0cc6abc..a7dc3b9 100644
--- a/backend/coldchain/src/process.rs
+++ b/backend/coldchain/src/process.rs
 
@@ -237,13 +237,8 @@ fn try_process_coldchain_notifications(
             ))
         })?;
 
-        let sensor_row = match sensor_row {
-            Some(row) => row,
-            None => {
-                log::error!("No sensor info found for sensor {}", sensor_id);
-                continue;
-            }
-        };
+        let sensor_row =
+            sensor_row.with_context(|| format!("No sensor info found for sensor {}", sensor_id))?;
 
         let alert = match sensor_status {
             SensorStatus::HighTemp => match config.high_temp {

```

_Originally posted by @andreievg in https://github.com/openmsupply/notify/pull/175#pullrequestreview-1675418567_

### LINKED ISSUE #213: Follow up reminders for cold chain alerts

Follow up reminders for cold chain alerts.

If an alert hasn't been resolved within the specified timeframe, send reminder alerts

image

## Acceptance Criteria

- [ ] If follow up reminders are disabled, no follow up alerts should be sent
- [ ] Otherwise follow up reminder should be sent at the configured interval until resolved
- [ ] Each sensor will get it's on OK message
- [ ] Duplicate configurations option `Send temperature OK confirmation` should be removed

jmbrunskill:
Closes #213 
Closes #199 

# 👩🏻‍💻 What does this PR do?

- [ ] Cold chain alerts now send reminders, with a reminder number
- [ ] Refactors the way we send cold chain alerts to align with suggestion from @andreievg (hopefully)
- [ ] Updates cold chain alert to use markdown. (Note that's actually not working properly, due to rendering issues with telegram)
- [ ] New unit tests for the refactored logic

Example...
image

I've added a bunch of tests to the code base and done some testing locally.
Would be good to try things like no-data on a longer timeframes I guess.

Reminders might get tedious if set too short, but setting them for 1 minute is good for testing :)

Telegram markdown doesn't work how you would expect it too.
e.g. you only put a single `*` for bold not `**` as per common mark format. This means it's not rendering well...
image
Will tackle this in #245


backend/coldchain/src/sensor_state.rs
```diff
@@ -5,12 +5,19 @@ use serde::{Deserialize, Serialize};
...
 use crate::ColdChainError;
 
-#[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Default)]
 pub struct SensorState {
     pub sensor_id: String,
     pub status: SensorStatus,
-    pub timestamp: NaiveDateTime,
+    #[serde(alias = "timestamp")]
+    pub timestamp_localtime: NaiveDateTime,
```
jmbrunskill, line 13: I changed from `timestamp` to `timestamp_local` to make it very clear that this isn't in UTC as that' confusing in the code right now. `#[serde(alias = "timestamp")]` should allow backwards compatibility for old records...

backend/coldchain/src/test/process.rs
```diff
@@ -0,0 +1,1086 @@
...
+    alerts::AlertType,
+    latest_temperature::LatestTemperatureRow,
+    parse::ColdChainPluginConfig,
+    process::{evaluate_sensor_status, try_process_sensor_notification},
+    sensor_info::SensorInfoRow,
+    sensor_state::{SensorState, SensorStatus},
+};
+
+#[test]
+fn test_evaluate_sensor_status() {
```
jmbrunskill, line 13: This test is unchanged from previous, just moved into this file

backend/coldchain/src/test/process.rs
```diff
@@ -0,0 +1,1086 @@
...
+        high_temp_threshold,
+        low_temp_threshold,
+        max_age,
+    );
+    // TODO: Old Data Logic https://github.com/openmsupply/notify/issues/179
+    assert_eq!(status, SensorStatus::NoData);
+}
+
+#[test]
+fn test_try_process_sensor_notification_prev_ok() {
```
jmbrunskill, line 131: `test_try_process_sensor_notification*` tests are all new

backend/service/src/notification_config/tests/duplicate.rs
```diff
@@ -11,9 +11,9 @@ mod notification_config_duplicate_tests {
     use std::sync::Arc;
 
     #[actix_rt::test]
-    async fn notification_config_service_update_success() {
+    async fn notification_config_duplicate_test() {
         let (_, _, connection_manager, _) = setup_all(
-            "notification_config_service_update_success",
+            "notification_config_duplicate_test",
```
jmbrunskill, line 16: This should be a bugfix on main...

backend/telegram/src/service/responder.rs
```diff
@@ -32,7 +32,7 @@ pub async fn handle_telegram_updates(
                 let chat_id = message.chat.id.to_string();
                 let text = message.text.unwrap_or_default();
 
-                if text.contains("/hello") || text.contains("/chat") {
+                if text.contains("/hello") || text.contains("/chat") || text.contains("/start") {
```
jmbrunskill, line 35: Small side track, allows the chatbot to respond when starting a direct message...

backend/templates/coldchain/no_data.md
```diff
@@ -0,0 +1,15 @@
+**Monitoring connection lost**
```
jmbrunskill, line 1: New wording from @adamdewey

backend/coldchain/src/process.rs
```diff
@@ -385,7 +255,200 @@ fn try_process_coldchain_notifications(
...
 }
 
-fn evaluate_sensor_status(
+pub fn try_process_sensor_notification(
+    config: &ColdChainPluginConfig,
+    prev_sensor_state: SensorState,
+    sensor_row: SensorInfoRow,
+    now_local: NaiveDateTime,
+    latest_temperature_row: Option<latest_temperature::LatestTemperatureRow>,
+) -> Result<(SensorState, Option<ColdchainAlert>), ColdChainError> {
```
andreievg, line 264: I don't think this returns ColdChainError ?

andreievg (approved):
Thank James,

As per our discussion:

Using base_alert may help with future refactor (including recent refactor with `last_data_time: last_data_localtime` which is not there for HighTemp, btw this is the only change request please, still approving though).

Internal database errors can be 'unwrapped' or propogated in a way that stops the server.

Serialisation/Deserialisation errors for sensor status can be treated similary as long as there is a test se/de for new shape and de test for previous data shape.

Can abstract the serialisation logic in a new method or at a service layer (i.e. at the time of consumption get_json vs get_value).

Pervious Sensor State as default (rather then None), when it doesn't exist, adds a few more use cases to think through.

Diagram update please (also good to capture this in issues).

Thanks, approving (just that one change please mentioned above)

backend/coldchain/src/process.rs
jmbrunskill: You're right, remove the error.

backend/coldchain/src/process.rs
```diff
@@ -167,97 +125,32 @@ fn try_process_coldchain_notifications(
                     "Failed to get previous state for sensor {}: {:?}",
                     sensor_id, e
                 ))
-            })?;
-
-        let prev_sensor_status = match prev_sensor_status {
-            Some(s) => SensorState::from_string(&s),
+            })?; // Exit the function if we get a database error...
+
+        let prev_sensor_state = match prev_sensor_state {
```
andreievg, line 130: A couple of carry over comments:

> Internal database errors can be 'unwrapped' or propogated in a way that stops the server.
> Serialisation/Deserialisation errors for sensor status can be treated similary as long as there is a test se/de for new shape and de test for previous data shape.

Which means should be able to return error here, if you still agree with that (rather the None of parsing failure). Otherwise if deserialisation fails, it's likely to occur again and previous state not recorded, causing an undefined state (which may lead to unexpected behaviour)

> Can abstract the serialisation logic in a new method or at a service layer (i.e. at the time of consumption get_json vs get_value).

Also similar to the above I think we also talked about this briefly

backend/coldchain/src/process.rs
```diff
@@ -274,73 +167,35 @@ fn try_process_coldchain_notifications(
...
-                    last_data_time: last_data_localtime,
-                    data_age,
-                    temperature: current_temp,
-                    alert_type: AlertType::Ok,
-                }),
-                false => {
-                    log::info!("Confirm Ok alert disabled for sensor {}", sensor_id);
-                    None
+        // if we have an updated state, persist it...
+        if prev_sensor_state.is_none() || sensor_state != prev_sensor_state.unwrap_or_default() {
```
andreievg, line 179: Btw is Some(&sensor_state) != prev_sensor_state.as_ref()

backend/coldchain/src/process.rs
```diff
@@ -274,73 +167,35 @@ fn try_process_coldchain_notifications(
...
-                    sensor_name: sensor_row.sensor_name.clone(),
-                    last_data_time: last_data_localtime,
-                    data_age,
-                    temperature: current_temp,
-                    alert_type: AlertType::NoData,
-                }),
-                false => {
-                    log::info!("No data alert disabled for sensor {}", sensor_id);
-                    None
+                Err(e) => {
```
andreievg, line 190: Another database error that could result in undefined state (can exist app)

PR #233, opened 2023-11-08
Allow notify to run on a different URL

### LINKED ISSUE #232: Allow notify to run on a different path
## Describe the bug

If we want to run notify behind a proxy server on a path such as /notify/ currently it doesn't work.

## To Reproduce

Setup a server such as (caddy)[https://caddyserver.com/download]

Using a Caddyfile config like this...
```
:7007 {
        respond "This request didn't make it to notify, sorry"
        redir /notify /notify/
        handle_path /notify/* {
                reverse_proxy http://localhost:8007
        }
}
```

1. Start Caddy and notify (on default port 8007)
2. go to http://localhost:7007/notify/
3. Notify doesn't load

## Expected behaviour

Should be able to run notify on this path (perhaps with a config change?)

## Additional context

This stack overflow suggests using a public path 
`publicPath: '/app'`
https://stackoverflow.com/questions/63674303/configure-webpack-to-serve-react-at-relative-path

Maybe we set it up to always run on /notify/?

jmbrunskill:
Closes #232

# 👩🏻‍💻 What does this PR do?
 

This makes it possible to run notify on a path like https://yourserver.yourdomain.com/notify/

- [ ] Uses webpack `publicPath = 'auto'` mode See: https://webpack.js.org/guides/public-path/#automatic-publicpath
- [ ] Uses webpack's `__webpack_public_path__` to calculate the graphql path, and `ReactRouter` `basename`
- [ ] Uses relative path for translation file loading

I've tried this with Caddy using a Caddy file like this.

```
:7007 {
	respond 404
	redir /notify /notify/
	handle_path /notify/* {
		reverse_proxy http://localhost:8007
	}
}
```

To serve the frontend via the backend server, you also need to run `yarn build`.

Anything else I might have missed?

### DIFF
--- a/frontend/packages/common/src/intl/context/IntlContext.tsx
+++ b/frontend/packages/common/src/intl/context/IntlContext.tsx
@@ -39,7 +39,7 @@ export const IntlProvider: FC<PropsWithChildrenOnly> = ({ children }) => {
             },
             {
               /* options for secondary backend */
-              loadPath: '/locales/{{lng}}/{{ns}}.json',
+              loadPath: 'locales/{{lng}}/{{ns}}.json',
             },
           ],
         },
--- a/frontend/packages/config/src/config.ts
+++ b/frontend/packages/config/src/config.ts
@@ -1,6 +1,8 @@
 declare const API_HOST: string;
 declare const APP_BUILD_VERSION: string;
 declare const BUGSNAG_API_KEY: string;
+/* eslint-disable camelcase */
+declare const __webpack_public_path__: string;
 
 // For production, API is on the same domain/ip and port as web app, available through sub-route
 // i.e. web app is on https://my.healthsupplyhub.com/, then graphql will be available https://my.openmsupply.com/graphql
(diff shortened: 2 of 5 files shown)


lache-melvin (commented):
Found this... I can navigate to any page, but refreshing only works when I'm not on a nested route? I.e. I can refresh fine on `/recipients`, but trying to refresh on `/recipients/sql`, I get this:

Screenshot 2023-11-22 at 9 33 44 AM

lache-melvin:
After some investigation into this, it looks like we'd need to choose between making the app available on a dynamic path (`publicPath: auto`), or using nested paths within the app. When loading the app on a nested route, without a predefined publicPath, Webpack's only option is to assume the app is being loaded on what it should consider the base path, so everything breaks.

Assuming we'd like to continue not constraining our app to a single level of routes, we could hardcode the publicPath to something other than `/`, as proposed in the issue, e.g. if `/notify` would better suit. Or we could inject the path as an environment variable. 

Be good to catch up on this/jam on what the best way forward would be @jmbrunskill 🙏

lache-melvin:
Update on this was that it is probably best to leave, sadly, couldn't find a solution that came without falling over somewhere else...

PR #239, opened 2023-11-09
Allow up to 1000 queries to be returned for selection

### LINKED ISSUE #237: Only 20 queries show in Notifications > Select Queries
## Describe the bug

In the notifications > Select Queries, only 20 queries show in the list and the filter works in these 20 queries only.
Screenshot 2023-11-09 at 12 36 00 PM

## To Reproduce

1. Create more than 20 queries  
2. Go to notifications > scheduled, click on "Select Queries"
3. Count the number of queries in the list
4. Only 20 shows even though you have more than 20 queries in Queries

jmbrunskill:
Closes #237

# 👩🏻‍💻 What does this PR do?
Quick hack to show all the queries

Added another issue for a longer term fix here:
https://github.com/msupply-foundation/notify/issues/238

Create more than 20 queries, check if you can select them or not!


mark-prins (approved):
👍 LGTM!

PR #248, opened 2023-11-15
246 add a run now button for notifications

### LINKED ISSUE #246: Add a `run now` button for notifications

When testing notifications, it would be very helpful to have a run now button which schedules the notification to run immediately.

Perhaps the the save button could be split to allow a save an run now option?
image

## Acceptance Criteria

- [ ] Easy to trigger a notification to run now. Maybe with a warning/confirmation modal?

A full featured testing system would be better in most instances, but this could still be useful even if that exists to push out an alert when something went wrong or if you updated a template found a problem and wanted to send out an update to all recipients?

jmbrunskill:
Closes #246 

# 👩🏻‍💻 What does this PR do?
 

Adds a run now button, which schedules the notification to run immediately

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -1719,6 +1719,7 @@ dependencies = [
  "async-graphql",
  "async-graphql-actix-web",
  "async-trait",
+ "chrono",
  "graphql_core",
  "graphql_types",
  "repository",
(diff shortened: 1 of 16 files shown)


lache-melvin (approved):
Works well for me James!

The one thing I picked up is that every time you click `Run` the entire edit page refreshes (only a loading indicator is rendered while the mutation is run.) I know in the background its a mutation, but it doesn't feel like one as a user, so the refresh is odd...

Quite possibly something just to capture in a separate issue, but would be much nicer to display the loading indicator next to the button as opposed to wiping the page - probably on `Save` as well as `Run`?

I think it would be great to get this in... as you said, at least we have an indication that something is happening!

PR #251, opened 2023-11-17
249 email error can stop processing

### LINKED ISSUE #249: Email error can stop processing?
## Describe the bug

It seems like notifications stopped processing, possibly after an email sending error?
image

```
SmtpError(lettre::transport::smtp::Error { kind: Permanent(Code { severity: PermanentNegativeCompletion, category: Unspecified3, detail: Five }), source: "Incorrect authentication data" })
```

## To Reproduce

1. Send an email with incorrect login info??

jmbrunskill:
Closes #249

Not sure if this really fixes the error we saw, as we don't have good logging, and I haven't been able to re-create it.
The closest I've come to is manually panicing in a plugin which does kill the scheduled task processing without stopping the whole server, so I've changed the behaviour so it does exit the server in this situation.

# 👩🏻‍💻 What does this PR do?

- [ ] Adds code to exit the server if a plugin, email, or other scheduled task process panics/unwraps/expects etc.
- [ ] Adds a timeout to lettre code, I think it already has this but just in case, added something, so it hopefullywon't hang forever

I added a failing unwrap to scheduled task processing, this closed the join handle and caused the whole server to shutdown.
```
2023-11-17 17:21:35.674677000 [INFO] <actix_server::server:196>:Actix runtime found; starting in Actix runtime
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', scheduled/src/lib.rs:37:27
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2023-11-17 17:21:35.739012000 [ERROR] <server:206>:Scheduled task stopped unexpectedly: Err(JoinError::Panic(Id(3), ...))
2023-11-17 17:21:35.739145000 [INFO] <actix_server::accept:143>:Accept thread stopped
2023-11-17 17:21:35.739271000 [INFO] <actix_server::worker:593>:Shutting down idle worker
2023-11-17 17:21:35.739354000 [INFO] <actix_server::worker:593>:Shutting down idle worker
2023-11-17 17:21:35.739406000 [INFO] <actix_server::worker:593>:Shutting down idle worker
2023-11-17 17:21:35.739483000 [INFO] <actix_server::worker:593>:Shutting down idle worker
2023-11-17 17:21:35.739591000 [INFO] <actix_server::worker:593>:Shutting down idle worker
2023-11-17 17:21:35.739661000 [INFO] <actix_server::worker:593>:Shutting down idle worker
2023-11-17 17:21:35.739745000 [INFO] <actix_server::worker:593>:Shutting down idle worker
2023-11-17 17:21:35.739850000 [INFO] <actix_server::worker:593>:Shutting down idle worker
2023-11-17 17:21:35.739917000 [INFO] <actix_server::worker:593>:Shutting down idle worker
2023-11-17 17:21:35.739972000 [INFO] <actix_server::worker:593>:Shutting down idle worker
2023-11-17 17:21:35.740138000 [INFO] <server:277>:Remote server stopped
```

We could also restart in this situation, but I'm thinking if we exit we aleast know something is wrong rather than going into an infinite restart loop.

Would be interested in a restart option, but would like to keep some kind of state so we only try to restart 5 times or something?

I tried to experiment with adding some kind of timeout to the plugin and notification sending calls.
Seems hard, would be good to peer program it with someone perhaps? Related to https://github.com/msupply-foundation/notify/issues/133

### DIFF
--- a/backend/server/src/lib.rs
+++ b/backend/server/src/lib.rs
@@ -202,10 +202,13 @@ async fn run_server(
         _ = ctrl_c => false,
         Some(_) = off_switch.recv() => false,
         _ = restart_switch_receiver.recv() => true,
+        scheduled_error = scheduled_task_handle => {
+            error!("Scheduled task stopped unexpectedly: {:?}", scheduled_error);
+            false
+        }
     };
 
     server_handle.stop(true).await;
-    scheduled_task_handle.abort();
     Ok(restart)
 }
 
--- a/backend/server/src/lib.rs
+++ b/backend/server/src/lib.rs
@@ -96,38 +96,42 @@ async fn run_server(
         Box::new(ColdChainPlugin::new()),
     ];
 
-    let scheduled_task_handle = actix_web::rt::spawn(async move {
+    let mut scheduled_task_handle = actix_web::rt::spawn(async move {
         scheduled_task_runner(scheduled_task_context, plugins).await;
     });
 
     // Setup a channel to receive telegram messages, which we want to handle in recipient service
     let telegram_token = config_settings.clone().telegram.token;
-    if let Some(telegram_token) = telegram_token {
-        let telegram_service = TelegramService::new(
-            TelegramClient::new(telegram_token),
-            config_settings.server.app_url.clone(),
-        );
-        let telegram_update_channel = telegram_service.init().await;
+    let telegram_update_handler_option = match telegram_token {
+        None => None,
+        Some(telegram_token) => {
+            let telegram_service = TelegramService::new(
+                TelegramClient::new(telegram_token),
+                config_settings.server.app_url.clone(),
+            );
+            let telegram_update_channel = telegram_service.init().await;
 
-        // Handle telegram updates in recipient service
-        let telegram_update_context =
-            ServiceContext::new(service_provider_data.clone().into_inner());
-        let telegram_update_context = match telegram_update_context {
-            Ok(telegram_update_context) => telegram_update_context,
-            Err(error) => {
-                return Err(std::io::Error::new(
-                    std::io::ErrorKind::Other,
-                    format!(
-                        "Error unable to create telegram update task context: {:?}",
-                        error
-                    ),
-                ));
-            }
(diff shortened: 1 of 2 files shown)


backend/server/src/lib.rs
```diff
@@ -202,10 +202,13 @@ async fn run_server(
...
         Some(_) = off_switch.recv() => false,
         _ = restart_switch_receiver.recv() => true,
+        scheduled_error = scheduled_task_handle => {
+            error!("Scheduled task stopped unexpectedly: {:?}", scheduled_error);
+            false
+        }
     };
 
     server_handle.stop(true).await;
-    scheduled_task_handle.abort();
```
clemens-msupply, line 208: Think you should still kill all running tasks. For example, if the server restarts. Or was there a reason to not do it?

In general, always keep track of all running tasks and stop them or join with them when finishing. For example, _telegram_update_handler is not joined. (From the docs: A JoinHandle detaches the associated task when it is dropped, which means that there is no longer any handle to the task, and no way to join on it.)

clemens-msupply: Ah you used it in the select!, don't know how to wait for it then. I don't select! is killing it

clemens-msupply: Sorry `&mut scheduled_task_handle` in select! should work

jmbrunskill: > For example, _telegram_update_handler is not joined. 

I tested it and on a shutdown just dropping the tasks seems fine, but you're right I don't think it will work correctly on a restart...
I'll add this, although I'm not sure when/why we'll need a restart. It's hangover from open-mSupply. Maybe it's better to just not have this feature at all?

clemens-msupply: Yes was wondering the same. This was also just an independent observation. Couldn't find a reason why the task would panic but think this is a good solution!

clemens-msupply (approved):
Looks good, lets hope it crashes the server next time :)

PR #254, opened 2023-11-19
Allow sorting for enabled configs

jmbrunskill:
Closes `nada`

# 👩🏻‍💻 What does this PR do?
 

Found it was useful to be able to sort notification config for ones that are enabled, especially when testing and you need to check lot of things

Check that the sorting works, and doesn't adversly affect anything else.


lache-melvin (approved):
Looks great 😎 

NIT: run a yarn generate so the graphql types are up to date 👍

PR #258, opened 2023-11-20
255 parameters in events table

### LINKED ISSUE #255: Show parameters in Events Table

The Tera Context, should be saved to the events table, to help with debugging

image

## Acceptance Criteria

- [ ] Easy to see exactly what was avaialbe or not in the rendering

lache-melvin:
Closes #255

# 👩🏻‍💻 What does this PR do?
 

Adds the parameters and query results as JSON to the bottom of the notification event log. Also adds the recipient name to the `To` field. This should help to improve troubleshooting.

Persists the tera context with the notification event during the render and enqueue step of the notification event.

Screenshot 2023-11-21 at 11 20 44 AM

- Trigger a new notification event - set a scheduled event for a minute from now, and wait for it to complete.
- View the event details for the new notification event
- See any parameters or query results in the JSON field at the bottom of the event details
- Recipient name is part of the `To` field

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2


lache-melvin:
Hm just seen a case where the recipient `name` and `to_address` are the same... is adding the name likely to be helpful, or just keep the to_address and remove this change?
Screenshot 2023-11-21 at 11 43 07 AM

frontend/packages/system/src/NotificationEvents/DetailView/DetailView.tsx
```diff
@@ -81,18 +86,16 @@ export const DetailView = () => {
             {entity?.errorMessage ? (
               <TextArea value={entity?.errorMessage} />
             ) : (
-              <Typography variant="body1">No Error</Typography>
+              <Typography variant="body1">{t('messages.no-error')}</Typography>
```
jmbrunskill, line 89: Thanks for doing these translations, I remember I was rushing at skipped this :)

frontend/packages/system/src/NotificationEvents/DetailView/eventContext.ts
```diff
@@ -0,0 +1,21 @@
+import { NotificationTypeNode } from '@common/types';
+
+export type EventContext = {
```
jmbrunskill, line 3: I'm not sure if this is strictly needed, we may want to use recipient data from the tera context, e.g 
```
Hi {{ recipient.name }},

Here's your report...
```

So I would be fine with removing this, the good thing is the whole context is in the DB so easy to change later, so maybe leave it as is for now...

jmbrunskill (approved):
Looks great. Thanks!

Added a new issue #262 which might need to take this further.
Still good to get this in though!

frontend/packages/system/src/NotificationEvents/DetailView/eventContext.ts
jmbrunskill: Yes I think it would be best to leave it unchanged

PR #259, opened 2023-11-20
253 fix escaping in logs

### LINKED ISSUE #253: `updateNotificationConfig` logging escaping keeps growing...
## Describe the bug

When updating a notification config, it appears the escaping from the previous config object isn't parsed, and so those escape characters are then escaped again on the next save...

Not a big issue, just clogging the logs:

Screenshot 2023-11-20 at 12 04 09 PM

## To Reproduce

1. Go to Notifications
2. Click on Create a new Scheduled notification
3. Update and save it several times
4. View logs for growing escapes

## Expected behaviour

Singular escapation :)

lache-melvin:
Closes #253

# 👩🏻‍💻 What does this PR do?
 

Turns out it wasn't a parsing/escaping issue... we weren't removing the old `configurationData` field from the config after it was parsed, so each time we saved the new `configurationData`, we included a child field of the previous `configurationData` - ah recursion!

- Create a new notification config, enable it, and click `Save and Run` a bunch of times
- View the log of the mutation input... configuration data should stay the same size :)

This doesn't fix the configuration data for all the configs that already have this nested stringified configuration data, but it does stop them growing any further. That should be fine right?

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/frontend/packages/system/src/Notifications/Pages/ColdChain/parseConfig.ts
+++ b/frontend/packages/system/src/Notifications/Pages/ColdChain/parseConfig.ts
@@ -14,10 +14,12 @@ export function parseColdChainNotificationConfig(
 ): CCNotification | null {
   if (!config) return null;
   try {
+    const { configurationData, ...rest } = config;
+
     return {
       ...defaultCCNotification,
-      ...JSON.parse(config.configurationData),
-      ...config,
+      ...JSON.parse(configurationData),
+      ...rest,
     };
   } catch (e) {
     showError();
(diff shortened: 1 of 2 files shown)


jmbrunskill (approved):
Looks good.
I think we need the same thing in Cold Chain?

https://github.com/msupply-foundation/notify/blob/b1949061d6cd7e8d6d6256bbd45038872356dc83/frontend/packages/system/src/Notifications/Pages/ColdChain/parseConfig.ts#L80

Slightly tempted to try to fix existing records but probably not worth the effort.

lache-melvin:
>  I think we need the same thing in Cold Chain?

Oh of course, thanks, will do that now :)

Issue #71, opened 2023-08-24
Enter key doesn't select checkboxes


lache-melvin:
## Describe the bug

You can use the tab key to tab through the Cold Chain notification toggle fields, but the enter key doesn't select them.
![Screenshot 2023-08-24 at 3 43 07 PM](https://github.com/openmsupply/notify/assets/55115239/a144420c-7e3a-44ac-ba72-0e71d0e34a70)

## To Reproduce

Steps to reproduce the behaviour:

1. Go to 'Notifications'
2. Click on 'New Notification'
3. Use the tab key to tab to a checkbox
4. Press enter
5. The checkbox doesn't select

## Expected behaviour

The enter key selects/deselects the highlighted checkbox

EthanMcQ-TMF:
The enter key _shouldn't_ affect checkboxes. The preferred web accessibility shortcut is spacebar and that already works:
https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/examples/checkbox/

PR #261, opened 2023-11-21
256 filter events by config

### LINKED ISSUE #256: filter events by notification config

Able to restrict events to a notification config

## Acceptance Criteria

- [ ] Link to recent events from the notification config page

lache-melvin:
Closes #256 

# 👩🏻‍💻 What does this PR do?
 

Adds a Filter by Notification Config button to the events list:
Screenshot 2023-11-21 at 5 02 30 PM

Clicking this button opens a list, where you can search for and select the config to filter by: 
Screenshot 2023-11-21 at 5 02 36 PM

This puts the config id in the url, filters the events, and replaces the Filter By button with the name of the config we are currently filtering on:
Screenshot 2023-11-21 at 5 02 18 PM

Opening the filter modal again, the Clear Filter button becomes available:
Screenshot 2023-11-22 at 9 10 52 AM

Also adds a button on the edit notification config page, linking to this filtered view of events for that config:
Screenshot 2023-11-21 at 5 01 18 PM

For some reason, any interaction with the keyboard while the modal is open causes the filter button to be in focus when it closes??? i.e. if you scroll and click, it's fine, but if you search and then click, this happens:

Screenshot 2023-11-21 at 4 57 42 PM

This seems very confusing to me, like I'm being summoned to open the modal again... I tried a couple of things but not sure why it's only when you use keyboard interactions. There is an option to disable the ripple effect for any keyboard focus on the button, but that's a loss for accessibility... which would be better for now??


jmbrunskill:
> This seems very confusing to me, like I'm being summoned to open the modal again... I tried a couple of things but not sure why it's only when you use keyboard interactions. There is an option to disable the ripple effect for any keyboard focus on the button, but that's a loss for accessibility... which would be better for now??

We also have this issue : https://github.com/msupply-foundation/notify/issues/71 is it related?

lache-melvin:
> We also have this issue : #71 is it related?

Mm quite possibly, I think we've got a few default behaviours getting confused with everything we've wrapped around the base MUI components 🥲

jmbrunskill (approved):
❤️ it!
Super useful.

It was immediately obvious to me that the Recent Events button goes on the top right.
It's consistent with the Notification Config link though, so like it from that perspective.

UX wise, it's super useful to have this feature so people will get used to it.

I'm open to ideas for re-arranging though. The event view feels a bit messy to, but again, it's for admins so they can learn what to click :)

lache-melvin:
> I'm open to ideas for re-arranging though. 

Yeah definitely agree it could use some UX design work, some of it isn't super intuitive at the moment 🥲 

I'll go with disabling the ripple effect, because you still _can_ select the button, and the ripple is quite obviously strange. I've created an issue to review it/the other issues with keyboard interactions... possibly a good mission for next RnD day!

PR #264, opened 2023-11-22
Tera playground

jmbrunskill:
RND Day

# 👩🏻‍💻 What does this PR do?
 

Uses wasm to run tera in the browser, and validate the templates the users entered

Kapture 2023-11-22 at 17 26 15

This probably needs a bit of testing, particulary for things like deployment.
Does it increase the bundle size too much, Can it be lazy loaded, etc...

- [ ] Probably need to do something to make sure the wasm package is built for production or dev builds, or at least users know they need too
- [ ] Maybe should use the proper error type which would involve https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.Error.html and/or https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.SyntaxError.html


frontend/packages/host/webpack.config.js
```diff
@@ -10,25 +10,28 @@ const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
...
+         * In "json" mode single JSON file with bundle report will be generated
+         */
+        analyzerMode: 'disabled',
+        generateStatsFile: true,
+      })
     : new DummyWebpackPlugin();
 
   return {
+    experiments: {
+      syncWebAssembly: true,
```
jmbrunskill, line 33: This is the `actual` change

jmbrunskill:
Results published as https://www.npmjs.com/package/@openmsupply/tera-web

Closing this PR

PR #267, opened 2023-11-23
245 telegram markdown is not rendered correctly

### LINKED ISSUE #230: Telegram doesn't handle emojis
## Describe the bug

Would be nice to be able to use emoji's like 🌡️  or 🔋 in telegram notifcations

Apparently we need to convert these to UNICODE or something?
https://stackoverflow.com/questions/31430587/how-to-send-emoji-with-telegram-bot-api

## To Reproduce

1. Create a scheduled alert going to telegram
1. Put emoji in template
1. See we get a telegram error

## Expected behaviour

Message should still send

### LINKED ISSUE #245: Telegram Markdown is not rendered correctly
## Describe the bug

image

## To Reproduce

1. Put in a markdown template such as **Title**
2. Send notification

## Expected behaviour

Title should be rendered in bold :)

## Additional context

Telegram is very particular, and specific about what it will accept.
https://core.telegram.org/bots/api#markdownv2-style

For example `*Title*` will be rendered in bold but `**Title**` will not, despite this being the default markdown style.

jmbrunskill:
Closes #245 
Closes #230 

# 👩🏻‍💻 What does this PR do?
 

- [ ] Parses common markdown format into something that will mostly work on for Telegram
- [ ] Adds emoji to cold chain templates (mainly to demonstrate this does actually work)

**BEFORE**
image

**AFTER**
image

Some test cases are in the code, And have done a bit of testing with cold chain and other formats.
Would be good to do more testing though!

There's a few telegram things like `~strikethrough~` and `||spoiler||` that are in telegram but not in common markdown.
Would be nice to not escape those things so that they correctly display if someone is writing telegram style markdown.
I'm thinking this can be a wait to see/hear what issues people face before doing more though.

BTW: This is already a bit of a time sink, and we'll never get a perfect representation of everything in telegram :(

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -3598,6 +3598,8 @@ version = "0.1.0"
 dependencies = [
  "http",
  "log",
+ "pulldown-cmark",
+ "regex",
  "reqwest",
  "serde",
  "serde_json",
--- a/backend/service/src/notification/mod.rs
+++ b/backend/service/src/notification/mod.rs
@@ -201,8 +201,13 @@ impl NotificationServiceTrait for NotificationService {
                 NotificationType::Telegram => {
                     // Try to send via telegram
                     if let Some(telegram) = &ctx.service_provider.telegram {
+                        let telegram_markdown_v2 =
+                            telegram::service::markdown::cmark_to_telegram_v2(
+                                &notification.message,
+                            );
+
                         let result = telegram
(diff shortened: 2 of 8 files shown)


lache-melvin (approved):
Looks good to me James, sounds like a good plan to wait to hear of any issues and invest any more time accordingly :)

Issue #262, opened 2023-11-21
Internal Error, Failed to add Template to Tera Context


jmbrunskill:
## Describe the bug
```
2023-11-22 09:30:02.411590000 [ERROR] <scheduled::process:40>:InternalError("InternalError(\"Failed to add body template to tera instance: Error { kind: Msg(\\\"Failed to parse 'body_template'\\\"), source: Some(Error { kind: Msg(\\\" --> 6:37\\\\n  |\\\\n6 | Of the {{ num_of_stores_configured.0. num_of_stores_configured }} stores configured: </br>␊\\\\n  |                                     ^---\\\\n  |\\\\n  = expected `or`, `and`, `not`, `<=`, `>=`, `<`, `>`, `==`, `!=`, `+`, `-`, `*`, `/`, `%`, a filter, or a variable end (`}}`)\\\"), source: None }) }\")")
```

## To Reproduce

Steps to reproduce the behaviour:

## Expected behaviour

Should these internal errors create an event that the user can see in the event log?
Or errors like this should be recorded directly against the Notification Config?

## Additional context

I think this might be related to #253 ? But need to check? If so does this mean we need to do something more to fix there?

jmbrunskill:
Maybe we should validate the template can be added to the tera context on save?

PR #270, opened 2023-11-23
262 show parsing/processing errors in event log

lache-melvin:
Part of #262

# 👩🏻‍💻 What does this PR do?
 

If there is an internal error during the processing of a notification event, create a phantom failed notification event with no recipient, so can see it in the notification event log.

Screenshot 2023-11-24 at 11 54 28 AM
Screenshot 2023-11-24 at 11 53 51 AM

Create and run a scheduled event with a broken body template. Go to the events page and see an empty failed event with the parsing error details.

- Seems a shame not to include the title if possible, but that may be the thing that failed to parse... probably possible but would cost more time
- The map_err is both a bit repetitive and I think not great that it has the side effect creating the event, but moving the adding of templates to tera to a separate `try_add_templates_to_tera` method involved a lot of passing things around e.g. the tera instance - but I'm not very confident in my Rust yet so maybe I'm just missing the way to do it!

### DIFF
--- a/backend/service/src/notification/enqueue.rs
+++ b/backend/service/src/notification/enqueue.rs
@@ -57,18 +57,6 @@ pub fn create_notification_events(
 ) -> Result<(), NotificationServiceError> {
     let repo = NotificationEventRowRepository::new(&ctx.connection);
 
-    let base_row = NotificationEventRow {
-        id: uuid(),
-        created_at: Utc::now().naive_utc(),
-        sent_at: None,
-        error_message: None,
-        send_attempts: 0,
-        updated_at: Utc::now().naive_utc(),
-        notification_config_id: config_id.clone(),
-        retry_at: None,
-        ..Default::default()
-    };
-
     // Dedup recipients by to_address
     let mut recipients = notification.recipients.clone();
     recipients.sort_by(|a, b| a.to_address.cmp(&b.to_address));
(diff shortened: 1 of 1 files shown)


backend/service/src/notification/enqueue.rs
```diff
@@ -287,4 +302,42 @@ mod test {
...
+        );
+
+        assert!(result.is_err());
+
+        // Check we have a notification event with error message
+        let notification_event_row_repository = NotificationEventRowRepository::new(&connection);
+        let notification_event_rows = notification_event_row_repository.un_sent().unwrap();
+
+        assert_eq!(notification_event_rows.len(), 1);
+        assert_eq!(notification_event_rows[0].to_address, "-".to_string());
```
jmbrunskill, line 336: Does this work? I Would have thought it would be default which would be empty string?

backend/service/src/notification/enqueue.rs
```diff
@@ -167,6 +154,34 @@ pub fn create_notification_events(
...
+        notification_type: NotificationType::Unknown,
+        ..Default::default()
+    };
+
+    match NotificationEventRowRepository::new(&ctx.connection)
+        .insert_one(&failed_notification_event_row)
+        .map_err(|e| NotificationServiceError::DatabaseError(e))
+    {
+        Ok(()) => NotificationServiceError::InternalError(format!(
+            "Failed to add template to tera instance: {:?}",
```
jmbrunskill, line 178: This feels a little bit too specific here? Could parameter error for example, is it body template or message template?

I guess you could pass through a string? Or just have a generic wording here?

jmbrunskill (approved):
I haven't had a chance to try it out, but looks good.
Couple of small comments...

I assume you've checked it doesn't crash the server or anything?
Merge when you're ready!

backend/service/src/notification/enqueue.rs
lache-melvin: Guess who made changes and didn't run the tests again 🙈 I blame jetlag 😂

PR #271, opened 2023-11-24
262 show template errors

### LINKED ISSUE #214: Scheduled Report templates don't take up the width of the the available screen
## Describe the bug

image

## To Reproduce

1. Open a scheduled report.

## Expected behaviour

The box should default to a wider view.

### LINKED ISSUE #262: Internal Error, Failed to add Template to Tera Context

lache-melvin:
Closes #262 (this PR is part 2)

# 👩🏻‍💻 What does this PR do?
 

Uses the tera-web package to validate whether the templates are parse-able, and displays an error state if not. Also prevents being able to save a broken template.

Screenshot 2023-11-24 at 2 04 40 PM

Makes subject and body template fields always full width too, to prevent their width jumping around as the error shows and hides: 
i.e. this closes #214 
Screenshot 2023-11-24 at 2 06 49 PM


frontend/packages/system/src/Notifications/Pages/Scheduled/ScheduledNotificationEditPage.tsx
```diff
@@ -55,9 +56,19 @@ export const ScheduledNotificationEditPage = () => {
...
+    } catch (e) {
+      return false;
+    }
+  };
+
   const isInvalid =
     !draft.title ||
-    // nothing selected
+    !isValidTemplate(draft.subjectTemplate) ||
+    !isValidTemplate(draft.bodyTemplate) ||
```
lache-melvin, line 71: Is it a valid notification with a subject but no body? Or the body/message is required also?

jmbrunskill: I think we should always have a body.
Telegram only renders the body, subject is just for email in current design.

jmbrunskill (approved):
Awesome! Love it.

PR #275, opened 2023-12-04
162 windows installer setup

### LINKED ISSUE #162: Windows installer

A windows installer similar to OMS

Installer should setup things like environment variables needed to correctly connect from rust to a postgres db
It should also install any default templates required

raviSussol:
Closes #162 

# 👩🏻‍💻 What does this PR do?
 
Added a windows installer build script for the notify app.

Download installer from [here](https://jenkins.msupply.org:8443/job/notifyMain%20-%20Installers) and test it.
Or, create a new tag (starts with vxx.xx.xx) and then push to the origin. Jenkins will automatically fire the windows installer build job in the https://jenkins.msupply.org:8443/job/notifyMain/

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2


build/windows/notify_service.suf
```diff
@@ -0,0 +1,4347 @@
...
+<Type>2</Type>
+</SessionVar>
+<SessionVar>
+<Name>%AppShortcutFolderName%</Name>
+<Value>%ProductName%</Value>
+<Type>2</Type>
+</SessionVar>
+<SessionVar>
+<Name>%notify_port%</Name>
+<Value>7000</Value>
```
jmbrunskill, line 4253:
```suggestion
<Value>7007</Value>
```


I've been using 7007 currently....

jmbrunskill (changes requested):
Great work!


However, It looks like the templates directory is missing?
Can you add this, and make sure it gets updated with an upgrade?
This is needed for the coldchain alerts, they use these templates.

image


Is it possible to get the build number from the git tag here?
image

I think we should probably use `Notify` instead of `Notify Service` (it's already a service)
image

Similar for the install location, I think we should probably use `C:\Program Files\notify`?
That does potentially create a problem for existing installs, but we can be across that when we update from the manual install to this installer...

jmbrunskill:
I also notice that if we're upgrading we still get the same prompts for passwords etc. Even these are already in the config file?
image

Is it possible to only ask these questions on the first install?

jmbrunskill (approved):
Thanks for the fast work Ravi!
Working great!

PR #276, opened 2023-12-06
Auto Backup sqlite

### LINKED ISSUE #174: Auto backup for sqlite file

For backups to work well, we should run something like `VACUUM INTO 'backup.sqlite'`  on the sqlite database, to integrate the WAL log and back a backup friendly file.

The user should be able to configure a backup interval and backup file that notify will automatically create on a schedule.

## Acceptance Criteria

- [ ] Configuration to turn on-off the backup
- [ ] Backup should be run on a schedule (specified by a cron)
- [ ] The backup path should be specified in the local.yaml file (the may want to put it on a different drive for example)
- [ ] Doesn't need to save old backups, it's essentially just a backup ready file

Manage `vacuum` in an external script

jmbrunskill:
Closes #174 

# 👩🏻‍💻 What does this PR do?
 
Runs a backup of the notify sqlite database on a schedule based on a cron job definition

Note: the cron format has extra fields to what I was used too
 sec  min   hour   day of month   month   day of week   year

The base.yaml has been setup with a default time of 9pm as requested by Nick.
We assume versioning an history of the backup can be managed by an external backup tool.

Would be good to make sure it can be turned off, and that the paths work correctly on windows

## 📃 Documentation

Should add some documentation around this


backend/server/src/auto_backup.rs
```diff
@@ -0,0 +1,66 @@
+use repository::backup_sqlite;
+use service::service_provider::ServiceContext;
+use std::path::Path;
+use tokio::time::sleep;
+
+use chrono::Utc;
```
raviSussol, line 6:
```suggestion
use chrono::offset::Utc;
```

raviSussol: Atm, cron datetime is in local but the `now_utc` is in utc.

raviSussol: Also, regarding to this timezone I guess this notify app is going to be installed next to the mSupply or open msupply app (i.e. on windows) so it would be a bit confusion or hassle to know and get a GMT (utc) time at the time of configuration? Wouldn't the local timezone is easy to use?

backend/repository/src/lib.rs
```diff
@@ -25,4 +26,13 @@ pub fn run_db_migrations(connection: &StorageConnection) -> Result<(), String> {
         .map_err(|err| format!("{}", err))
 }
 
+pub fn backup_sqlite(connection: &StorageConnection, path: &str) -> Result<(), String> {
+    log::info!("Backing up sqlite database to {}", path);
+    sql_query("VACUUM INTO ?")
```
raviSussol, line 31: This vaccum into query expects the backup path to be existed before running this query but the backup folder path doesn't get created before. Would be very helpful if get create automatically (if not exist)?

raviSussol (commented):
@jmbrunskill Nice feature! Changes look good. Though a query regarding backup timezone and a suggestion.

raviSussol:
Plus I think we also need to include this backup settings in the windows installer?

backend/repository/src/lib.rs
jmbrunskill: Yes good idea, I'll look at doing that tomorrow.

backend/server/src/auto_backup.rs
jmbrunskill: Yes it would be better in local time. I'll look into this more closely.
Very good pick up! I didn't think about the timezones!

jmbrunskill:
> Plus I think we also need to include this backup settings in the windows installer?

I think the defaults will be picked up, but yes it wouldn't hurt to update the installer too...

raviSussol (approved):
Looks great now!

jmbrunskill:
> Looks great now!

Thanks Ravi! I'll do a new build.

PR #280, opened 2023-12-12
fix search bar clearance race condition

lache-melvin:
[As found in Universal Codes](https://github.com/msupply-foundation/unified-codes/pull/434#discussion_r1423310098)

# 👩🏻‍💻 What does this PR do?
 

There was some kind of race condition going on with the clear search `X` button. Sometimes the `onClick` would fire and clear the search, but sometimes the focus would shift away from the input first, so the search field would shrink without clearing.

This PR updates the clear logic to run on `onMouseDown`, which fires before the focus change, so the search will reliably clear!

Type in a search field and clear it lots of times 😅


jmbrunskill (approved):
Works great thanks!

PR #288, opened 2024-02-14
285 generate parameters from sql

### LINKED ISSUE #285: Generate parameters from SQL query

It should be possible to run an SQL Query to get the parameters to run a scheduled notification for.

For example if a new Masterlist (AKA Project) is added, we'd like to be able to add this to the notification parameters automatically.

The user should be able to specify a query that will define the parameters for the notification to run with.
This allow the system to automatically respond to changes in database configuration, for example if you want a notification to run for every mSupply site, when a new site it added, the system should automatically include it the next time the notification runs.

## Acceptance Criteria

- [ ] Parameters can be specified either as a Query, or manually.

Manually run a query and copy and paste in when changes happen.

EthanMcQ-TMF:
Closes #285

# 👩🏻‍💻 What does this PR do?
 

Added a new section to notifications to add an SQL query. Currently these are the same queries used for notifications

Screenshot 2024-02-15 at 12 32 45

~A parameter query is expected to return a single row with a column "parameters", as a JSON array in the same format as our JSON parameter editing in the UI~

**A parameter query is expected to return a column for each parameter, with each row being used as a parameter set**

Screenshot 2024-02-16 at 09 50 16

When you run the notification, a new event should generate for each defined parameter from the query result (In this example I'm just listing all the tables in my local database and the id of the first entry (There's a number of failures because I don't have data in all my tables)

Screenshot 2024-02-15 at 12 38 14

**This should work alongside regularly defined parameters!** The list of query generated parameters and the list of user provided parameters will be merged when creating events.

In addition I've made it possible for a notification to be defined with **no** user defined parameters, so a notification can be created without any parameter sets (and pull exclusively from query parameters). I've added an additional "create" button for adding a new parameter when none exists already

Tested locally running the queries screenshotted above ⬆️ 

This is working, but there's definitely some improvements we could make. Currently I'm just using notification queries for fetching parameters, splitting those out to their own defined table similar to sql recipient lists is an obvious improvement.

This is also my first time writing rust so let me know if I've done anything silly there 😅 

## 📃 Documentation

- [ ] Notification Parameters - Get parameters from SQL query
- [ ]


frontend/packages/system/src/Notifications/Pages/Base/NotificationDetailPanel.tsx
```diff
@@ -112,10 +139,25 @@ export const NotificationDetailPanel = ({
...
+                    onUpdateParams(idx, key, param[key] || '');
+                  } else {
+                    // If the key exists on the parameter object but not the JSON, delete it
+                    onDeleteParam(idx, key);
+                  }
                 }
               });
+              while (params.length > editedParams.length) {
+                onDeleteParam(editedParams.length, null);
+              }
```
EthanMcQ-TMF, line 160: This whole section of code is completely separate to the main issue, it just improves the behaviour of loading params from JSON. The previous implementation didn't allow for things like removing parameter sets, deleting individual keys from parameter sets, or adding empty parameter sets

frontend/packages/system/src/Notifications/Pages/Scheduled/ScheduledNotificationEditPage.tsx
```diff
@@ -44,6 +44,10 @@ export const ScheduledNotificationEditPage = () => {
       (entity as NotificationConfigRowFragment) ?? null,
       parsingErrorSnack
     );
+    if (parsedDraft) {
+      // Backend expects this value to be null if unedited so don't load it from config
+      parsedDraft.nextDueDatetime = null;
+    }
```
EthanMcQ-TMF, line 50: Not associated with main issue but fixes a bug. The current config of `nextDueDatetime` would always be sent alongside the rest of the data when sending an update to the config, when the backend sees this is the update query, it runs the notification. So the notification would always run on save whether you hit "Save" or "Save & Run"

backend/service/src/notification_config/parameters.rs
```diff
@@ -0,0 +1,80 @@
...
+            ))
+        })?;
+    let mut sql_params: Vec<HashMap<String, serde_json::Value>>  = serde_json::from_str(&sql_params_string)
+        .map_err(|e| {
+            NotificationServiceError::InternalError(format!(
+                "Failed to parse notification sql parameters (expecting an array of params_string): {:?} - {}",
+                e, params_string
+            ))
+        })?;
+    all_params.append(&mut sql_params);
```
jmbrunskill, line 37: This can be written like this instead.
I'm not sure if one is strictly better than the other but in general I try to only make this mutable if they have to be.
```suggestion
    let sql_params: Vec<HashMap<String, serde_json::Value>>  = serde_json::from_str(&sql_params_string)
        .map_err(|e| {
            NotificationServiceError::InternalError(format!(
                "Failed to parse notification sql parameters (expecting an array of params_string): {:?} - {}",
                e, params_string
            ))
        })?;
    all_params.extend(sql_params);
```
Weird to me that append requires `&mut`, but I guess it needs the same datatype...

I think it's fine how it is but wanted to check why i needed to be `mut`

backend/service/src/notification_config/parameters.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    return Ok(all_params);
+}
+
+fn get_sql_parameters(
+    ctx: &ServiceContext,
+    parameter_query_id: &String,
+) -> Result<String, NotificationServiceError> {
+    // TODO: Maybe split these to a new database table
```
jmbrunskill, line 46: It's quite a bit of work, and I'm not sure it's needed but could be handy, or maybe add a query type so we can filter?

backend/service/src/notification_config/parameters.rs
```diff
@@ -0,0 +1,80 @@
...
+        })?;
+
+    let parsed_results: Vec<HashMap<String, serde_json::Value>> = serde_json::from_str(&query_result.results)
+        .map_err(|e| {
+            NotificationServiceError::InternalError(format!("Failed to parse parameter query results: {:?}", e))
+        })?;
+
+    let parameters = match parsed_results[0].get("parameters") {
+        None => return Err(NotificationServiceError::InternalError(format!(
+                    "No 'parameters' column found in query results - {}", parameter_query_id)
```
jmbrunskill, line 74: If we did create a new method we could automatically run something like `SELECT json_agg(q.*) as parameters from (<THEQUERY>) as q`?

Then we wouldn't have to the restrictions about the column naming and parameter queries would be a bit easier to test?

backend/service/src/notification_config/update.rs
```diff
@@ -124,5 +126,8 @@ pub fn generate(
     // Note: We usually reset the next check datetime in case the schedule has changed, or something needs to be recalculated
     new_notification_config_row.next_due_datetime = next_due_datetime;
 
+    // We might want to clear out the parameter_query_id so it's allowed to be None
```
jmbrunskill, line 129: I guess we're just seeing to empty string now? It seems to be working??

frontend/packages/system/src/Notifications/Pages/Base/NotificationDetailPanel.tsx
```diff
@@ -37,48 +44,68 @@ export const NotificationDetailPanel = ({
     params = [params];
   }
 
-  if (params.length === 0 || params[0] === undefined) {
-    params = [{} as KeyedParams];
-  }
+  const { queryParams } = useQueryParamsState();
+  const { data: queriesData } = useNotificationQueries(queryParams);
+  const selectedQuery = queriesData?.nodes.find(query => query.id === parameterQueryId);
```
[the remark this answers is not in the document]
jmbrunskill: I'm ok with grabing every thing for now.
I also didn't expect 500 parameter sets,so you never know...

frontend/packages/system/src/Notifications/Pages/Base/NotificationDetailPanel.tsx
```diff
@@ -112,10 +139,25 @@ export const NotificationDetailPanel = ({
...
+                  params.push({});
+                  onDeleteParam(idx, 'this-is-a-hack-to-force-an-update');
+                }
+                // Iterate through all valid keys
+                for (const key of [...Object.keys(param), ...Object.keys(params[idx] || {})]) {
+                  if (param[key]) {
+                    // If we parsed the key from the JSON input, update it
+                    onUpdateParams(idx, key, param[key] || '');
+                  } else {
+                    // If the key exists on the parameter object but not the JSON, delete it
```
jmbrunskill, line 153: Appreciate all these comments!

frontend/packages/system/src/Notifications/Pages/Scheduled/ScheduledNotificationEditPage.tsx
jmbrunskill: Nice catch

jmbrunskill (approved):
Fantastic work! 

Happy to merge this, but also happy if you wanted to add something to smooth the creation of the parameter queries.
Would be nice if we didn't have to have a column called `parameters`.
I think it would be better if just have a list of rows, each row becomes a parameter set? That would basically need to call `pg_sql_query_as_json_rows`  I think?

backend/service/src/notification_config/update.rs
EthanMcQ-TMF: This was more a comment about why it's not using the `if let Some(X)` check we use for other values. Since that doesn't let you un-set it back to `None`

EthanMcQ-TMF:
> Fantastic work!
> 
> Happy to merge this, but also happy if you wanted to add something to smooth the creation of the parameter queries. Would be nice if we didn't have to have a column called `parameters`. I think it would be better if just have a list of rows, each row becomes a parameter set? That would basically need to call `pg_sql_query_as_json_rows` I think?

Nah much easier than that, the results are already json-d, I just removed the section that expects a column called `parameters` and it works perfectly fine

Screenshot 2024-02-16 at 09 50 16

jmbrunskill:
> Nah much easier than that, the results are already json-d, I just removed the section that expects a column called `parameters` and it works perfectly fine

Nice! Was thinking that after I added that comment, isn't it already json...

PR #290, opened 2024-02-27
Don't repeat duplicate query parameters

### LINKED ISSUE #289: Reusing parameters in Tera template causes duplicate parameter edit entries
## Describe the bug

If a parameter is used twice in the same notification, the edit dialog will have duplicates

Screenshot 2024-02-27 at 15 57 08

## To Reproduce

1. Create a notification query
2. Use the same parameter twice e.g. `SELECT {{table}}.id FROM {{table}}`

## Expected behaviour

Each parameter should only have one edit field

## Additional context

I'm just writing this so I can put a PR against it

EthanMcQ-TMF:
Closes #289

# 👩🏻‍💻 What does this PR do?
 
Don't return duplicates when parsing Tera parameters

I've added an additional unit test for this case

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2

### DIFF
--- a/frontend/packages/common/src/utils/tera/TeraUtils.test.ts
+++ b/frontend/packages/common/src/utils/tera/TeraUtils.test.ts
@@ -16,6 +16,11 @@ describe('TeraUtils', () => {
       'data.name',
     ]);
   });
+  it('Should not extract duplicate parameters', () => {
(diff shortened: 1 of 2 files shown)


jmbrunskill (approved):
Looks good to me!

PR #292, opened 2024-03-01
266 - Fixup tool tip errors

### LINKED ISSUE #266: React forward ref errors with Tooltip
## Describe the bug

I'm getting React warnings on both the events list and event details pages, about an issue with the forward refs for some `Tooltip`s...

Event list: 
Screenshot 2023-11-23 at 9 09 32 AM

Event details:
Screenshot 2023-11-23 at 9 09 44 AM

## To Reproduce

## Expected behaviour

No errors

EthanMcQ-TMF:
Closes #266

# 👩🏻‍💻 What does this PR do?
 
Fix console errors that occur when displaying certain tooltips.

Loading up the events view and checking the console, you should no longer see tooltip errors.

These errors were also causing the tooltips to fail to display, so you should now be able to hover over date entries and actually see the tooltip

Screenshot 2024-03-01 at 16 41 43

Screenshot 2024-03-01 at 16 45 44

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/frontend/packages/system/src/NotificationEvents/DetailView/DetailView.tsx
+++ b/frontend/packages/system/src/NotificationEvents/DetailView/DetailView.tsx
@@ -53,20 +53,22 @@ export const DetailView = () => {
         {/* if we have a config_id, create a link to edit the config */}
         {entity?.notificationConfigId && (
           <Tooltip title={entity.notificationConfig?.title ?? ''}>
-            <BaseButton
-              onClick={() => {
-                navigate(
-                  configRoute(
-                    entity.notificationConfig?.kind ?? ConfigKind.Scheduled,
-                    entity.notificationConfigId ?? ''
-                  )
-                );
-              }}
-              variant="outlined"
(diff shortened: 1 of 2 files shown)


frontend/packages/system/src/NotificationEvents/DetailView/DetailView.tsx
```diff
@@ -53,20 +53,22 @@ export const DetailView = () => {
...
-                  )
-                );
-              }}
-              variant="outlined"
-              endIcon={<EditIcon />}
-            >
-              {t('button.edit-notification-config')}
-            </BaseButton>
+            <div>
+              <BaseButton
```
EthanMcQ-TMF, line 57: Mui tooltip expects it's child to accept `ref` which react function components do not. A simple fix is to just... wrap it in a non-function component

jmbrunskill (approved):
Thanks! I didn't realise the reason but fixed something similar recently.
Wrapping in a non-functional component seems to be a good quick fix.

PR #293, opened 2024-03-11
Actually use "retry_at" field in notifications

### LINKED ISSUE #92: backoff for notification retries

EthanMcQ-TMF:
Closes #92

# 👩🏻‍💻 What does this PR do?
 
Rather than retrying errored notifications immediately, wait 15/30/60 minutes (double each retry)

Tested by intentionally erroring a notification (easy to do, just don't run an instance of mailhog and all notifs will error), tested with a reduced `RETRY_DELAY_MINUTES` of `1`

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/backend/repository/src/db_diesel/notification_event_row.rs
+++ b/backend/repository/src/db_diesel/notification_event_row.rs
@@ -109,7 +109,8 @@ impl<'a> NotificationEventRowRepository<'a> {
             .filter(
                 notification_event_dsl::status
                     .eq(NotificationEventStatus::Queued)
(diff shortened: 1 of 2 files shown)


jmbrunskill (approved):
Nice, very clean and simple!

PR #294, opened 2024-07-04
hacky conditional scheduled tasks

jmbrunskill:
Closes #

# 👩🏻‍💻 What does this PR do?
 

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2

### DIFF
--- a/backend/scheduled/src/parse.rs
+++ b/backend/scheduled/src/parse.rs
@@ -1,4 +1,4 @@
-use chrono::{DateTime, Days, Months, Utc};
+use chrono::{DateTime, Days, Duration, Months, Utc};
 use serde::{Deserialize, Serialize};
 
 use crate::NotificationError;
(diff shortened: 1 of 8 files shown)


backend/scheduled/src/parse.rs
```diff
@@ -68,6 +72,17 @@ impl ScheduledNotificationPluginConfig {
         // Then add the duration to the schedule_start_time until we're past now
 
         match self.schedule_frequency.as_str() {
+            "hourly" => {
```
jmbrunskill, line 75: Sounds like the use case we're looking at first will be daily backups. So we could drop hourly if we think it could cause problems..

EthanMcQ-TMF:
In the realm of hacky solutions, in theory couldn't we put a conditional on a recipients query so the notification only has recipients when it needs to be sent out?

jmbrunskill:
> In the realm of hacky solutions, in theory couldn't we put a conditional on a recipients query so the notification only has recipients when it needs to be sent out?

That is a nice hacky idea, might give it a try...

jmbrunskill:
@EthanMcQ-TMF in your opinion what would be needed to make it less hacky?

PR #296, opened 2024-11-04
295 Fixup date truncation

### LINKED ISSUE #295: "End of month" schedules will break after a short month
## Describe the bug

The way we currently calculate the next due date for a monthly schedule will always cause truncation.

Currently we take the start of the schedule and loop through adding months until we reach a date that is beyond the current date. This will always (eventually) cause a schedule at the end of the month to truncate lower.

```
2024-01-31 + 1 month =>
2024-02-29 + 1 month => // rust::chrono automatically truncates to the smaller month
2024-03-29 // This should be the 31st
```

## To Reproduce

1.  Create a monthly scheduled notification for the last day of a 31 day month
2.  Wait

## Expected behaviour

The notification should send on the last day of the month, each month

## Additional context

This can also cause issues with "first of the month" schedules if the local timezone is ahead of UTC (e.g. NZT +12/+13) as the schedule will be processed on the server in UTC and this same truncation will happen

EthanMcQ-TMF:
Closes #295

# 👩🏻‍💻 What does this PR do?
 
- Calculate monthly schedule all at once to prevent truncation

- Add a unit test for monthly configs running through short months

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/backend/scheduled/src/parse.rs
+++ b/backend/scheduled/src/parse.rs
@@ -91,15 +91,17 @@ impl ScheduledNotificationPluginConfig {
                 return Ok(next_due_date);
             }
             "monthly" => {
+                let mut look_forward = 1;
                 let mut next_due_date = self.schedule_start_time;
                 while next_due_date < now_utc {
                     // Note: chrono automatically handles leap years and returns last day of the month if day isn't valid for that month
                     // https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDateTime.html#method.checked_add_months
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
Thanks Ethan!

I can't help but feel like it's crying out for an optimisation to get the difference in days/hours/months and then adjust based on that,b ut can definitely be in a future PR, and the original goal was to keep this logic simple so I like that you've kept it consistent.

PR #299, opened 2025-04-11
Additional notes on Notify README

rfromthecastle:
Added some information on the README to assist with issues that may arise when setting up and running Notify locally. Also added a few more steps about repository cloning for those who are not very familiar with GitHub.


jmbrunskill (approved):
Sounds like a good change

PR #301, opened 2025-10-03
300 required querys

### LINKED ISSUE #300: Allow rule-setting for sending notifications

In some cases for the scheduled notifications, we may only want to be notified if certain conditions are met or certain thresholds are passed, for example, a notification would only be sent when there is one or more store that hasn't synced in the last 90 days. Currently, there is no way to specify a rule for this behaviour.

Allow the option to set rules for when scheduled notifications should be triggered, in addition to the time schedule.

## Acceptance Criteria

- [ ] For scheduled notifications, provide a section to define rules that must be met before sending a scheduled notification.

Still send a notification in cases where the rules are not met, but use an if-block in the message template to display a message such as 'No issues to report'.

jmbrunskill:
Closes #300

# 👩🏻‍💻 What does this PR do?
Makes it possible to mark a query as required.
If a query is required scheduled notifications won't run when there's no data returned for that query.

I think there's some kind of bug where the notification config gets over-written when clicking required.
I saw it happen twice but can't replicate now.

## 📃 Documentation

_No user facing changes_

or

- [ ] Required Query Functionality...

### DIFF
--- a/backend/scheduled/src/process.rs
+++ b/backend/scheduled/src/process.rs
@@ -4,7 +4,10 @@ use chrono::{DateTime, NaiveDateTime, Utc};
 use repository::{NotificationConfigKind, NotificationConfigRowRepository};
 use service::{
     notification::enqueue::{create_notification_events, NotificationContext, TemplateDefinition},
-    notification_config::{query::NotificationConfig, recipients::get_notification_targets, parameters::get_notification_parameters},
+    notification_config::{
+        parameters::get_notification_parameters, query::NotificationConfig,
+        recipients::get_notification_targets,
+    },
     service_provider::ServiceContext,
 };
 
@@ -121,7 +124,12 @@ fn try_process_scheduled_notifications(
     let param_results = get_notification_parameters(ctx, &scheduled_notification);
     let mut all_params = match param_results {
         Ok(val) => val,
-        Err(e) => return Err(NotificationError::InternalError(format!("Failed to fetch parameters: {:?}", e)))
+        Err(e) => {
+            return Err(NotificationError::InternalError(format!(
+                "Failed to fetch parameters: {:?}",
+                e
+            )))
+        }
     };
 
     if all_params.len() == 0 {
@@ -153,7 +161,21 @@ fn try_process_scheduled_notifications(
             continue;
         }
 
-        let sql_query_parameters = get_notification_query_results(ctx, sql_params, &config)?;
+        let sql_query_parameters = get_notification_query_results(
+            ctx,
+            sql_params,
+            &config,
+            config.required_query_ids.clone(),
+        )?;
+
+        // If any required queries were skipped, skip this notification
+        let sql_query_parameters = match sql_query_parameters {
+            crate::query::NotificationQueryResult::Success(results) => results,
+            crate::query::NotificationQueryResult::Skipped(reason) => {
+                log::info!("Skipping notification: {}", reason);
+                continue;
+            }
+        };
 
         // Template data should include the notification config parameters, plus the results of any queries
         template_params.extend(sql_query_parameters);
--- a/backend/scheduled/src/process.rs
+++ b/backend/scheduled/src/process.rs
@@ -29,7 +29,8 @@ pub fn process_scheduled_notifications(
         .find_all_due_by_kind(ctx, NotificationConfigKind::Scheduled, current_time)
         .map_err(|e| NotificationError::InternalError(format!("{:?}", e)))?;
     let notifications_processed = scheduled_notifications.len();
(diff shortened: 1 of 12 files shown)


backend/scheduled/src/query.rs
```diff
@@ -47,6 +53,17 @@ pub fn get_notification_query_results(
                 json!([{"error": "error running query", "query": query.query, "parameters": parameters}])
             }
         };
+
+        // If the query is required and there are no results, return an error
+        if required_query_ids.contains(&query.id)
+            && (query_json.as_array().map_or(true, |arr| arr.is_empty()))
```
jmbrunskill, line 59: This was copilot, but I think readable enough?

frontend/packages/system/src/Notifications/Pages/Scheduled/ScheduledNotificationEditForm.tsx
```diff
@@ -119,6 +119,11 @@ export const ScheduledNotificationEditForm = ({
           setSelection={props => {
             onUpdate(props as Partial<ScheduledNotification>);
           }}
+          requiredQueryIds={draft.requiredQueryIds}
+          setRequiredQueryIds={ids => {
+            const patch = { ...draft, requiredQueryIds: ids };
```
jmbrunskill, line 124: I don't need draft here do I, maybe that's where I'm somehow overwriting state?

frontend/packages/system/src/Notifications/components/SqlQuerySelector.tsx
```diff
@@ -52,19 +70,35 @@ export const SqlQuerySelector: FC<QueryListProps> = ({
       label: 'label.query',
       width: 150,
       sortable: false,
-      accessor: ({ rowData }) => StringUtils.ellipsis(rowData?.query, 100),
+      accessor: ({ rowData }) => StringUtils.ellipsis(rowData?.query, 50),
```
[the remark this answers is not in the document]
jmbrunskill: This was to create more space on the screen for required.

backend/scheduled/src/process.rs
```diff
@@ -153,7 +161,21 @@ fn try_process_scheduled_notifications(
...
+            sql_params,
+            &config,
+            config.required_query_ids.clone(),
+        )?;
+
+        // If any required queries were skipped, skip this notification
+        let sql_query_parameters = match sql_query_parameters {
+            crate::query::NotificationQueryResult::Success(results) => results,
+            crate::query::NotificationQueryResult::Skipped(reason) => {
+                log::info!("Skipping notification: {}", reason);
```
lache-melvin, line 175: I would have expected this to result in a `skipped: 1` ?

Screenshot 2025-10-03 at 5 09 05 PM

frontend/packages/system/src/Notifications/Pages/Scheduled/ScheduledNotificationEditForm.tsx
lache-melvin: Yeah had a play, its working here - I think some strange stale state is hanging around with the datatable...

This fixes both the initial disappearing, and also to be able to select multiple queries as required:

https://github.com/msupply-foundation/notify/compare/300-required-querys...300-required-querys(state-fresh)

backend/scheduled/src/process.rs
jmbrunskill: Yeah I did think that would be nice. Will see if I can find a nice (& quick) way to do it.

jmbrunskill:
> When I was testing from the UI, honestly I was expecting to see it show up in the `Events` list? But if that's more work, all good, in reality you're probably much more interested in when we do send notifications than when we don't...

I think showing every skipped notification event in the list will potentially create a lot more logs & data than we want. When testing I think it would be helpful but once its up and running I think it's too much, although happy to be convinced otherwise.

jmbrunskill:
> > I think showing every skipped notification event in the list will potentially create a lot more logs & data than we want. When testing I think it would be helpful but once its up and running I think it's too much, although happy to be convinced otherwise.
> 
> Yeah - seeing the results of `save and run` would be great, even it it's usually not included in the events... I think ideally we'd actually see those results in the configuration page. Definitely sounding like a someday issue though!

Yeah that would be great, but a lot of work :)

jmbrunskill:
> Not sure if your reply meant you're still looking into mapping this to a "skipped" rather than "processed" result, or if that was harder than it was worth - either way happy for this to go in when ready 🚀

I think I've found a nice solution... 

[d825d5e](https://github.com/msupply-foundation/notify/pull/301/commits/d825d5ed25c6b5d851e7978297aacac8affc725a)

PR #313, opened 2026-04-29
updated the fulldate as yesterday's date

sthamanisha:
Closes #

# 👩🏻‍💻 What does this PR do?
 The out-of-stock logic was previously returning items that were unavailable at any point prior to the current date, rather than specifically identifying those that were out of stock as of yesterday. 
Now, fulldate is compared with yesterday's date with stock 0 to get out of stock on yesterday's date.

- [ ] Out-of-stock notification only lists items with SOH = 0
- [ ] The notification includes only the items that are out of stock on previous day's date

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2

### DIFF
--- a/docs/reference_notifications.md
+++ b/docs/reference_notifications.md
@@ -154,72 +154,74 @@ Sent every day, this notification lists all stock items that are completely out 
 #### Data Query — `stock_out_today`
 
 ```sql
- with pos AS(
-  SELECT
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
Looks great thanks @sthamanisha !

